PR #4, opened 2021-05-25
8267695: Bump update version for OpenJDK: jdk-11.0.13

GoeLin:
Hi,

this is the version bump for the 11.0.13 release.
This will be the first change directly pushed to the 
git repository for jdk11u-dev.

Obvioulsy, I will only push this after tagging 11.0.13+0, 
and after the git transition of the jdk11u-dev repo.


GoeLin:
Ok, seems I must redo this later :)

PR #22, opened 2021-06-10
8251945: SIGSEGV in PackageEntry::purge_qualified_exports()

zhengyu123:
I would like to fix the crash in openjdk 11u.

The crash is caused by racy installing new CLD in ClassLoaderDataGraph::add_to_graph().

The method first creates new ClassLoaderData, and in its constructor, it creates unnamed module entry and installs it in java_lang_Module oop.

Then add_to_graph() tries to install newly created CLD to java_lang_ClassLoader oop via CAS. If it loses race, then it deletes new CLD and returns existing one.

But at this point, java_lang_Module oop still points module entry that is already freed.

The fix I am purposing is to borrow ClassLoaderDataGraph_lock from JDK-8210155, but only uses it to prevent racing installing CLD and new CLD is still published via CAS to avoid needing additional patches.

Test:
 - [x] hotspot_runtime
 - [x] hotspot_gc
 - [x] vmTestbase_vm_gc

---------
### Progress

### Issue
 * [JDK-8251945](https://bugs.openjdk.java.net/browse/JDK-8251945): SIGSEGV in PackageEntry::purge_qualified_exports()

### Reviewers
 * [Andrew Dinn](https://openjdk.java.net/census#adinn) (@adinn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/22/head:pull/22` \
`$ git checkout pull/22`

`$ git checkout pull/22` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/22/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 22`

`$ git pr show -t 22`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/22.diff">https://git.openjdk.java.net/jdk11u-dev/pull/22.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/javaClasses.cpp
+++ b/src/hotspot/share/classfile/javaClasses.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -4020,7 +4020,7 @@ int  java_lang_ClassLoader::name_offset = -1;
 int  java_lang_ClassLoader::nameAndId_offset = -1;
 int  java_lang_ClassLoader::unnamedModule_offset = -1;
 
-ClassLoaderData* java_lang_ClassLoader::loader_data(oop loader) {
+ClassLoaderData* java_lang_ClassLoader::loader_data_acquire(oop loader) {
   assert(loader != NULL && oopDesc::is_oop(loader), "loader must be oop");
   return HeapAccess<MO_ACQUIRE>::load_at(loader, _loader_data_offset);
 }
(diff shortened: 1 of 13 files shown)


jerboaa:
Closed in favour of #23

src/hotspot/share/classfile/javaClasses.cpp
```diff
@@ -4022,12 +4022,18 @@ int  java_lang_ClassLoader::unnamedModule_offset = -1;
 
 ClassLoaderData* java_lang_ClassLoader::loader_data(oop loader) {
```
adinn, line 4023: I understand the desire to minimize changes to related files (specifically the JFR files that currently include calls to java_lang_ClassLoader::loader_data() and java_lang_ClassLoader::set_loader_data(). However, I think it risks misleading future maintainers to leave these methods with their existing names when they are actually being modified to employ acquire/release semantics.

I think it would be much better to rename them consistently with the upstream patch so they are called java_lang_ClassLoader::loader_data_acquire() and java_lang_ClassLoader::release_set_loader_data() and then work those name changes through to all client calls. That not only retains consistency with upstream it also clarifies the local use.

adinn (changes requested):
This looks fine apart from the naming of the get and set methods.

src/hotspot/share/classfile/javaClasses.cpp
zhengyu123: @adinn Thanks for reviewing. 

Make sense, updated accordingly.

adinn (approved):

PR #72, opened 2021-06-25
8269415: [11u] Remove ea from DEFAULT_PROMOTED_VERSION_PRE in OpenJDK 11u

RealCLanger:
Implement suggested parameter correction as per https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2021-February/005045.html

---------
### Progress

### Issue
 * [JDK-8269415](https://bugs.openjdk.java.net/browse/JDK-8269415): [11u] Remove ea from DEFAULT_PROMOTED_VERSION_PRE in OpenJDK 11u

### Reviewers
 * [Severin Gehwolf](https://openjdk.java.net/census#sgehwolf) (@jerboaa - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/72/head:pull/72` \
`$ git checkout pull/72`

`$ git checkout pull/72` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/72/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 72`

`$ git pr show -t 72`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/72.diff">https://git.openjdk.java.net/jdk11u-dev/pull/72.diff</a>

</details>


jerboaa (commented):
I'm not sure this is the right move. 11.0.13 (which corresponds to the jdk11u-dev tree) is "Early Access" for all intents and purposes at this time. It should be removed for the `-ga` tag only and then added back if you'd ask me.

At least that's how I understood Andrew's suggestion on the mailing list.

RealCLanger:
> I'm not sure this is the right move. 11.0.13 (which corresponds to the jdk11u-dev tree) is "Early Access" for all intents and purposes at this time. It should be removed for the `-ga` tag only and then added back if you'd ask me.

Well, my motivation for this change is mainly consistency. For instance in [16u](https://github.com/openjdk/jdk16u/blob/master/make/autoconf/version-numbers) you also find 'DEFAULT_PROMOTED_VERSION_PRE=', without it being changed back and forth in its history. We have never changed such type of release data in the source code so far and I think it's neither feasible nor the right thing to do, given that we use tags for releases.

I'm inclined to read this value like 'ea' designates the major release to be still in development.

After all, this property is only used in Oracle CI builds so far.

jerboaa:
> > I'm not sure this is the right move. 11.0.13 (which corresponds to the jdk11u-dev tree) is "Early Access" for all intents and purposes at this time. It should be removed for the `-ga` tag only and then added back if you'd ask me.
> 
> Well, my motivation for this change is mainly consistency. For instance in [16u](https://github.com/openjdk/jdk16u/blob/master/make/autoconf/version-numbers) you also find 'DEFAULT_PROMOTED_VERSION_PRE=', without it being changed back and forth in its history. We have never changed such type of release data in the source code so far and I think it's neither feasible nor the right thing to do, given that we use tags for releases.
> 
> I'm inclined to read this value like 'ea' designates the major release to be still in development.

Could be. I now also realize that it's the default for **PROMOTED** builds, which I suspect will never be EA for update releases.

> After all, this property is only used in Oracle CI builds so far.

In the bug you say:

```
While the value itself is only used by Oracle builds so far, it should still be corrected in OpenJDK to enable downstream vendors to correctly use it. 
```
This implies to me that we are going to maintain this to be correct going forward. Go from `ea` to `ga` at the correct points in time. Could you please clarify this in the bug?

RealCLanger:
> This implies to me that we are going to maintain this to be correct going forward. Go from `ea` to `ga` at the correct points in time. Could you please clarify this in the bug?

I've updated the bug. Please check whether the description is concise now.

jerboaa (approved):

PR #109, opened 2021-07-07
8269661: JNI_GetStringCritical does not lock char array

zhengyu123:
I would like to backport this patch to openjdk11u.

This patch mainly impacts Shenandoah GC, which is the only GC that supports object pinning and string deduplication.

The original patch does not apply cleanly, because it uses new string deduplication API introduced in jdk17. openjdk11u cannot prevent a string from being deduplicated, so it risks the possibility that string deduplication modifies string value while string critical session is in progress.

I purpose to take an alternative approach in openjdk11u: return a copy of string value when object pinning + string deduplication is enabled.

Test:
- [x] tier1 with UseShenandoahGC + UseStringDeduplication on Linux x86_64.

---------
### Progress

### Issues
 * [JDK-8269661](https://bugs.openjdk.java.net/browse/JDK-8269661): JNI_GetStringCritical does not lock char array
 * [JDK-8269650](https://bugs.openjdk.java.net/browse/JDK-8269650): Optimize gc-locker in [Get|Release]StringCritical for latin string

### Reviewers
 * [Aleksey Shipilev](https://openjdk.java.net/census#shade) (@shipilev - **Reviewer**) ⚠️ Review applies to ac6da0262f4af2bc91a69133a1eb187b54f02f09

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/109/head:pull/109` \
`$ git checkout pull/109`

`$ git checkout pull/109` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/109/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 109`

`$ git pr show -t 109`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/109.diff">https://git.openjdk.java.net/jdk11u-dev/pull/109.diff</a>

</details>


src/hotspot/share/prims/jni.cpp
```diff
@@ -3212,30 +3213,63 @@ JNI_ENTRY(void, jni_ReleasePrimitiveArrayCritical(JNIEnv *env, jarray array, voi
...
 JNI_END
 
+// If a copy of string value should be returned instead
+static bool copy_string_value(oop str) {
+  return java_lang_String::is_latin1(str) ||
+         // To prevent deduplication from replacing the value array while setting up or in
+         // the critical section. That would lead to the release operation
+         // unpinning the wrong object.
+         (Universe::heap()->supports_object_pinning() &&
+          StringDedup::is_enabled());
```
shipilev, line 3223: The indenting here is odd, should probably be 2 spaces?

src/hotspot/share/prims/jni.cpp
```diff
@@ -3212,30 +3213,63 @@ JNI_ENTRY(void, jni_ReleasePrimitiveArrayCritical(JNIEnv *env, jarray array, voi
 HOTSPOT_JNI_RELEASEPRIMITIVEARRAYCRITICAL_RETURN();
 JNI_END
 
+// If a copy of string value should be returned instead
+static bool copy_string_value(oop str) {
```
shipilev, line 3217: I propose we call this `should_copy_string_value`.

shipilev (commented):
This looks okay to me, modulo minor comments. But I also think we need to wait for upstream to test the patches first.

src/hotspot/share/prims/jni.cpp
```diff
@@ -3212,30 +3213,63 @@ JNI_ENTRY(void, jni_ReleasePrimitiveArrayCritical(JNIEnv *env, jarray array, voi
...
 JNI_END
 
+// If a copy of string value should be returned instead
+static bool should_copy_string_value(oop str) {
+  return java_lang_String::is_latin1(str) ||
+    // To prevent deduplication from replacing the value array while setting up or in
+    // the critical section. That would lead to the release operation
+    // unpinning the wrong object.
+    (Universe::heap()->supports_object_pinning() &&
+     StringDedup::is_enabled());
```
shipilev, line 3223: This could be on the same line, I think.

src/hotspot/share/prims/jni.cpp
```diff
@@ -3212,30 +3213,63 @@ JNI_ENTRY(void, jni_ReleasePrimitiveArrayCritical(JNIEnv *env, jarray array, voi
...
-  oop s = lock_gc_or_pin_object(thread, string);
-  typeArrayOop s_value = java_lang_String::value(s);
-  bool is_latin1 = java_lang_String::is_latin1(s);
-  if (isCopy != NULL) {
-    *isCopy = is_latin1 ? JNI_TRUE : JNI_FALSE;
-  }
+  oop s = JNIHandles::resolve_non_null(string);
   jchar* ret;
-  if (!is_latin1) {
+  if (!should_copy_string_value(s)) {
```
shipilev, line 3251: Can probably swap the branches to avoid `!` here? This would also be consistent with release.

shipilev (approved):
This looks good to me, but I still think we need to wait for more testing of these patches upstream.

PR #205, opened 2021-08-06
8251549: Update docs on building for Git

jerboaa:
Hi,

Now that OpenJDK 11u is fully on Git we should update the docs to no longer refer to Mercurial and use Git instead. The JDK 16 patch doesn't apply clean due to some differences in the file (e.g. referring to JDK 8 vs. JDK 11 in head). `doc/building.html` is the result of running `make update-build-docs` when the `building.md` file has the relevant changes.

Thoughts?

---------
### Progress

### Issue
 * [JDK-8251549](https://bugs.openjdk.java.net/browse/JDK-8251549): Update docs on building for Git

### Reviewers
 * [Christoph Langer](https://openjdk.java.net/census#clanger) (@RealCLanger - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/205/head:pull/205` \
`$ git checkout pull/205`

`$ git checkout pull/205` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/205/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 205`

`$ git pr show -t 205`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/205.diff">https://git.openjdk.java.net/jdk11u-dev/pull/205.diff</a>

</details>

### DIFF
--- a/doc/building.html
+++ b/doc/building.html
@@ -93,12 +93,10 @@
 <li><a href="#getting-help">Getting Help</a></li>
 </ul></li>
 <li><a href="#hints-and-suggestions-for-advanced-users">Hints and Suggestions for Advanced Users</a><ul>
-<li><a href="#setting-up-a-forest-for-pushing-changes-defpath">Setting Up a Forest for Pushing Changes (defpath)</a></li>
 <li><a href="#bash-completion">Bash Completion</a></li>
 <li><a href="#using-multiple-configurations">Using Multiple Configurations</a></li>
 <li><a href="#handling-reconfigurations">Handling Reconfigurations</a></li>
 <li><a href="#using-fine-grained-make-targets">Using Fine-Grained Make Targets</a></li>
-<li><a href="#learn-about-mercurial">Learn About Mercurial</a></li>
 </ul></li>
 <li><a href="#understanding-the-build-system">Understanding the Build System</a><ul>
 <li><a href="#configurations">Configurations</a></li>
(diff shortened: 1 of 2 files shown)


shipilev (commented):
The `*.md` change looks good, but `*.html` has a lot of irrelevant changes. This is usually caused by a different version of pandoc used to generate the HTML file. Please run `make update-build-docs` on master and see what changes it brings. The usual way is to find the pandoc version that produces the minimal changes to HTML. If we don't do this, then HTML would change every time a future contributor changes MD again.

jerboaa:
> The `*.md` change looks good, but `*.html` has a lot of irrelevant changes. This is usually caused by a different version of pandoc used to generate the HTML file.

I see. Makes sense. Do you know which pandoc version we need? I have `2.9.2.1`

> Please run `make update-build-docs` on master and see what changes it brings. The usual way is to find the pandoc version that produces the minimal changes to HTML. If we don't do this, then HTML would change every time a future contributor changes MD again.

OK. I'll try to figure it out. Thanks for the sanity check!

shipilev:
> > The `*.md` change looks good, but `*.html` has a lot of irrelevant changes. This is usually caused by a different version of pandoc used to generate the HTML file.
> 
> I see. Makes sense. Do you know which pandoc version we need? I have `2.9.2.1`

Look at `./make/conf/jib-profiles.js` in the 11u tree and in current master. Chances are there is no exact Pandoc version that produces the same HTML file _exactly_, but we probably want to have it as close to current version as possible. So that we generate the HTML file once here, and then stick with the particular Pandoc version going forward.

jerboaa:
> > > The `*.md` change looks good, but `*.html` has a lot of irrelevant changes. This is usually caused by a different version of pandoc used to generate the HTML file.
> > 
> > 
> > I see. Makes sense. Do you know which pandoc version we need? I have `2.9.2.1`
> 
> Look at `./make/conf/jib-profiles.js` in the 11u tree and in current master. Chances are there is no exact Pandoc version that produces the same HTML file _exactly_, but we probably want to have it as close to current version as possible. So that we generate the HTML file once here, and then stick with the particular Pandoc version going forward.

Thanks!

Found the needed version, I think. Seems to be `1.17.2`
https://github.com/openjdk/jdk11u-dev/blob/master/make/devkit/createPandocBundle.sh#L34

jerboaa:
@shade Updated `doc/building.html` with with pandoc `1.17.2` now. Looks clean-ish to me. Thoughts?

shipilev (approved):
Thanks! There are a few unrelated hunks still (formatting in cross-compilation tables), but I don't think we need to fret about those. Looks good.

jerboaa:
Thanks for the review!

RealCLanger (approved):
LGTM, thanks for backporting.

PR #279, opened 2021-08-23
8263776: [JVMCI] add helper to perform Java upcalls

jerboaa:
Here is a backport for JDK-8263776. The patch doesn't apply cleanly since it uses features only introduced with JDK-8230199 (JDK 15+, not in OpenJDK 11u). This patch is the same labsopenjdk 11 uses. This patch is also needed for using graal js with plain OpenJDK.

Testing: test/hotspot/jtreg/compiler/jvmci and manual upstream Graal unit test.

---------
### Progress

### Issue
 * [JDK-8263776](https://bugs.openjdk.java.net/browse/JDK-8263776): [JVMCI] add helper to perform Java upcalls

### Reviewers

### Contributors
 * Tom Rodriguez `<never@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/279/head:pull/279` \
`$ git checkout pull/279`

`$ git checkout pull/279` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/279/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 279`

`$ git pr show -t 279`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/279.diff">https://git.openjdk.java.net/jdk11u-dev/pull/279.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/jvmci/jvmciRuntime.cpp
+++ b/src/hotspot/share/jvmci/jvmciRuntime.cpp
@@ -415,7 +415,7 @@ class ArgumentPusher : public SignatureIterator {
   jlong _argument;
   bool _pushed;
 
-  jlong next_arg(BasicType expectedType) {
+  jlong next_arg() {
     guarantee(!_pushed, "one argument");
     _pushed = true;
     return _argument;
@@ -444,7 +444,7 @@ class ArgumentPusher : public SignatureIterator {
   }
 
  public:
-  ArgumentPusher(Symbol* signature, JavaCallArguments*  jca, jlong argument, bool is_static) : SignatureIterator(signature) {
+  ArgumentPusher(Symbol* signature, JavaCallArguments*  jca, jlong argument) : SignatureIterator(signature) {
     this->_return_type = T_ILLEGAL;
     _jca = jca;
     _argument = argument;
@@ -454,13 +454,13 @@ class ArgumentPusher : public SignatureIterator {
 
   inline void do_object() { _jca->push_oop(next_object()); }
 
-  inline void do_bool()   { if (!is_return_type()) _jca->push_int((jboolean) next_arg(T_BOOLEAN)); }
-  inline void do_char()   { if (!is_return_type()) _jca->push_int((jchar) next_arg(T_CHAR)); }
-  inline void do_short()  { if (!is_return_type()) _jca->push_int((jint)  next_arg(T_SHORT)); }
-  inline void do_byte()   { if (!is_return_type()) _jca->push_int((jbyte) next_arg(T_BYTE)); }
-  inline void do_int()    { if (!is_return_type()) _jca->push_int((jint)  next_arg(T_INT)); }
+  inline void do_bool()   { if (!is_return_type()) _jca->push_int((jboolean) next_arg()); }
+  inline void do_char()   { if (!is_return_type()) _jca->push_int((jchar) next_arg()); }
+  inline void do_short()  { if (!is_return_type()) _jca->push_int((jint)  next_arg()); }
+  inline void do_byte()   { if (!is_return_type()) _jca->push_int((jbyte) next_arg()); }
+  inline void do_int()    { if (!is_return_type()) _jca->push_int((jint)  next_arg()); }
 
-  inline void do_long()   { if (!is_return_type()) _jca->push_long((jlong) next_arg(T_LONG)); }
+  inline void do_long()   { if (!is_return_type()) _jca->push_long((jlong) next_arg()); }
   inline void do_float()  { if (!is_return_type()) _jca->push_float(next_float()); }
   inline void do_double() { if (!is_return_type()) _jca->push_double(next_double()); }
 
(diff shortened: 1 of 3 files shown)


jerboaa:
I'm pretty sure the GHA tier1 failure of `gc/stress/gcold/TestGCOldWithCMS.java` on Linux x86 is unrelated as this is an JVMCI only change.

src/hotspot/share/jvmci/jvmciRuntime.cpp
```diff
@@ -409,6 +409,115 @@ JRT_ENTRY(void, JVMCIRuntime::throw_class_cast_exception(JavaThread* thread, con
...
+      case T_SHORT:
+        return (jshort) value->i;
+      case T_INT:
+        return value->i;
+      case T_LONG:
+        return value->j;
+      case T_FLOAT:
+        return value->f;
+      case T_DOUBLE:
+        return value->d;
```
shipilev, line 513: The original commit uses `value->i` and `value->j` here. Of course, `value->f` and `value->d` look more correct, but do you understand why this difference?

src/hotspot/share/jvmci/jvmciRuntime.cpp
```diff
@@ -409,6 +409,115 @@ JRT_ENTRY(void, JVMCIRuntime::throw_class_cast_exception(JavaThread* thread, con
...
+ public:
+  ArgumentPusher(Symbol* signature, JavaCallArguments*  jca, jlong argument, bool is_static) : SignatureIterator(signature) {
+    this->_return_type = T_ILLEGAL;
+    _jca = jca;
+    _argument = argument;
+    _pushed = false;
+    iterate();
+  }
+
+  inline void do_object() {       _jca->push_oop(next_object()); }
```
shipilev, line 455: Excess whitespace before `_jca`.

src/hotspot/share/jvmci/jvmciRuntime.cpp
```diff
@@ -409,6 +409,115 @@ JRT_ENTRY(void, JVMCIRuntime::throw_class_cast_exception(JavaThread* thread, con
...
+  inline void do_int()    { if (!is_return_type()) _jca->push_int((jint)  next_arg(T_INT)); }
+
+  inline void do_long()   { if (!is_return_type()) _jca->push_long((jlong) next_arg(T_LONG)); }
+  inline void do_float()  { if (!is_return_type()) _jca->push_float(next_float()); }
+  inline void do_double() { if (!is_return_type()) _jca->push_double(next_double()); }
+
+  inline void do_object(int begin, int end) { if (!is_return_type()) do_object(); }
+  inline void do_array(int begin, int end)  { if (!is_return_type()) do_object(); }
+
+  inline void do_void()                     { }
```
shipilev, line 470: Excess whitespace before `{ }`.

shipilev (commented):
I have questions :)

shipilev:
> I'm pretty sure the GHA tier1 failure of `gc/stress/gcold/TestGCOldWithCMS.java` on Linux x86 is unrelated as this is an JVMCI only change.

Yes, it looks like another instance of [JDK-8272195](https://bugs.openjdk.java.net/browse/JDK-8272195).

src/hotspot/share/jvmci/jvmciRuntime.cpp
jerboaa: The original change handles `T_INT` and `T_FLOAT` (same bit width) with a single case statement. Same for `T_DOUBLE` and `T_LONG`. Therefore returning the more generic union types `value->i` and `value->j` respectively. I'm not sure this actually makes any difference. This seems the cleaner way to handle it.

shipilev: Yeah, I don't see the difference, but I think this is how we do conversions: put the `jdouble` to `value->d`, read `jlong` from `value->j`. Since this method returns `jlong`, I would expect `value->j` here. See e.g. `next_double()` in this patch.

Note that if we do `->f`, `->d` for `T_FLOAT` and `T_DOUBLE`, then it is reasonable to expect `T_BOOLEAN` and friends also replaced by their own type-specific union fields.

I would personally like to match mainline here, and then maybe clean up mainline, then backport the cleanup as the follow-up.

jerboaa: Sure. I've also noticed that there are minor differences to the JDK 17 version (local `return_type` variable over `jap.get_ret_type()`). Will fix. Thanks!

jerboaa:
@shipilev I've updated the patch to be similar to the mainline version now. Changes include:

- Get rid of unused `is_static` boolean in `ArgumentPusher`
- Get rid of unused `expectedType` in `nextArg()`
- Use local variable `return_type` over calling `jap.get_ret_type()` repeatedly.
- Use the same fall-through case for `T_DOUBLE/T_LONG` (`value->j`) and `T_INT/T_FLOAT` (`value->i`), which matches mainline.

shipilev (approved):
All right, this looks good to me.

jerboaa:
Thanks for the review!

PR #274, opened 2021-08-23
8250810: Push missing parts of JDK-8248817

rnkovacs:
This is a refactoring patch with no functional change. It didn't apply cleanly. Differences to the original commit:

In `thread_windows_x86.cpp`:
- Some removed lines were simpler on tip because of [JDK-8248240](https://github.com/openjdk/jdk/commit/a6e794b3e85d28537bf740d9d7e68339707d38e1#diff-d4111d540c34b1e0f4acac1e92877ce458b29d70126ccb113b439261a503ed95).

In `os_windows.cpp`:
- The block around L2461 is no longer present on tip, that and a bit of the next block can be simplified.
- The conditions on L2539 ([JDK-8234372](https://github.com/openjdk/jdk/commit/4e4d1f2b4dad61b4c4590f1d3a956b9ad86ff74c#diff-84938e7dfe64805a850b8a729495117a972a474622fff12bb97ee45cbc2184f2)) and L2549 ([JDK-8213199](https://github.com/openjdk/jdk/commit/7c3f2b06f13e3d28b74193d8acac4478e8c12bdb#diff-84938e7dfe64805a850b8a729495117a972a474622fff12bb97ee45cbc2184f2)) (original line numbers) are different on tip. We actually didn't touch those, they only show up colored in the diff because of the whitespace changes.
- The surrounding code is different around L2586 because of [JDK-8191278](https://github.com/openjdk/jdk/commit/aedbb7580334cb9131f547845f9b7aa3916ddf4b#diff-84938e7dfe64805a850b8a729495117a972a474622fff12bb97ee45cbc2184f2), but the changes are the same.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8250810](https://bugs.openjdk.java.net/browse/JDK-8250810): Push missing parts of JDK-8248817

### Reviewers
 * [Andrew Haley](https://openjdk.java.net/census#aph) (@theRealAph - **Reviewer**) ⚠️ Review applies to a06cf59aeec047e02492ba76397a9cfbf4dd4677
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Contributors
 * Ludovic Henry `<luhenry@openjdk.org>`
 * Bernhard Urban-Forster `<burban@openjdk.org>`

### Reviewing
<details><summary>Using <code>git</code></summary>

[...]

PR #300, opened 2021-08-29
Backport 257809d7440e87ac595d03b6c9a98d8f457f314c

rnkovacs:
Not a clean backport. Differences to the original commit:

- The removed lines in `stubGenerator_x86_32.cpp` and `stubGenerator_x86_64.cpp` are slightly different on tip (comments - [JDK-8234563](https://bugs.openjdk.java.net/browse/JDK-8234563), MP checks - [JDK-8188764](https://bugs.openjdk.java.net/browse/JDK-8188764)), but fundamentally the same.
- Changes to `add`, `xchg` and `cmpxchg` definitions in `atomic_windows_x86.hpp` needed to be adjusted to keep the original parameter orders (changed in JDK 14 by [JDK-8234563](https://bugs.openjdk.java.net/browse/JDK-8234563)).

Changes to `os_windows_x86.cpp` and `os_windows_x86.hpp` applied cleanly.

This is part of the Windows/AArch64 port.


lewurm (approved):
Looks good to me (but I'm not a formal reviewer).

theRealAph (commented):
Is it truly necessary to make such extensive changes to x86 code? I can understand why it's necessary to do such refactoring in head, but can't Windows x86 be mostly left alone, with Windows AArch64 added?

TheRealMDoerr (approved):
Changing so much x86 code in an Windows aarch64 preparation change is not so nice, but it's better than integrating half changes IMHO. Looks good and all tests are green. Good work!

theRealAph:
> Changing so much x86 code in an Windows aarch64 preparation change is not so nice, but it's better than integrating half changes IMHO.

Understood, but one of the main reasons new ports are easily accepted into maintenance releases of OpenJDK is that they are pure additions, so there's little churn to existing code. Backports should be minimal, in order to minimize that churn, and I don't believe this one is.

RealCLanger:
> > I think technically this one is reviewed but maintainers now need to take the decision whether to allow this or reject as being to much of churn. I suggest you add the fix request data to JBS and we will then take a decision.
> 
> It's not technically reviewed: even if two people approve it, objections must be addressed.
> 
> I am still of the belief that the AArch64 code can be integrated without touching the x86 atomics. If I had a WIndows AArch64 system and enough time to figure out how to build and debug this stuff I'd do it myself, just to prove the point.
> 
> Hint: put the common atomic stuff in a separate place. Call it from the AArch64 code, but do not touch the x86 code.

OK, I somehow overread the technical suggestion to try to find a way to do the changes only in the aarch64 code and leaving x86 alone. So, yes, if that would be possible it would probably be a more acceptable way of doing this backport.

lewurm:
By coincidence, the relevant parts for the Windows/AArch64 backport are in https://github.com/openjdk/jdk11u-dev/pull/274 anyway. So I agree with Andrew, this PR with the x86 changes are not needed for the Windows/AArch64 backport.

PR #373, opened 2021-09-17
8222446: assert(C->env()->system_dictionary_modification_counter_changed()) failed: Must invalidate if TypeFuncs differ

TheRealMDoerr:
Backport of JDK-8222446. Doesn't apply cleanly. Mainly because JVMCI is different, but also other context changes.

---------
### Progress

### Issue
 * [JDK-8222446](https://bugs.openjdk.java.net/browse/JDK-8222446): assert(C->env()->system_dictionary_modification_counter_changed()) failed: Must invalidate if TypeFuncs differ

### Reviewers
 * [Paul Hohensee](https://openjdk.java.net/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/373/head:pull/373` \
`$ git checkout pull/373`

`$ git checkout pull/373` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/373/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 373`

`$ git pr show -t 373`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/373.diff">https://git.openjdk.java.net/jdk11u-dev/pull/373.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -95,7 +95,7 @@ static bool firstEnv = true;
 
 // ------------------------------------------------------------------
 // ciEnv::ciEnv
-ciEnv::ciEnv(CompileTask* task, int system_dictionary_modification_counter)
+ciEnv::ciEnv(CompileTask* task)
   : _ciEnv_arena(mtCompiler) {
   VM_ENTRY_MARK;
 
@@ -115,7 +115,6 @@ ciEnv::ciEnv(CompileTask* task, int system_dictionary_modification_counter)
   assert(!firstEnv, "not initialized properly");
 #endif /* !PRODUCT */
 
-  _system_dictionary_modification_counter = system_dictionary_modification_counter;
   _num_inlined_bytecodes = 0;
   assert(task == NULL || thread->task() == task, "sanity");
   if (task != NULL) {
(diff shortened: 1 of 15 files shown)


phohensee (changes requested):
Missing copyright date updates in ciEnv.hpp, systemDictionary.*, dependencies.hpp,  jvmciEnv.*, method.cpp, callGenerator.cpp, parse1.cpp, jvmtiRedefineClasses.cpp, and vmStructs.cpp. Otherwise lgtm.

TheRealMDoerr:
@phohensee Thank you for reviewing my backport! The original change doesn't modify the Copyright headers in these files. So, I don't think my backport shouldn't do that. That would create further conflicts for future backports. We usually only change Copyright headers in backport PRs when the original change does so.

phohensee:
My understanding is that it's a legal requirement (but, ianal), to update a file's copyright date to that of the latest patch that affected it, regardless of the effect on future backports. In this case, the reason the original patch didn't update the copyright is because the not-affected files were already at 2019 when the original patch was applied. But, best to ask on the list, I suppose.

GoeLin:
Hi Paul, 
so far we never updated the Copyrights. If there are conflicts, we resolve to the latest date.
But if there are no conflicts, or if we have to resolve the file and the file has no Copyright change
in the original, or if the copyright is changed but the new year is in the past, we do not adapt it. 
Oracle handles this similarly.
Best regards,
  Goetz

phohensee (approved):
If that's the case (and in any case!) a note should be added to the updates wiki.

TheRealMDoerr:
Thanks again for the review! Right, would be good to have that documented. I couldn't even find official Copyright year update rules for the jdk project.
/integrate

PR #410, opened 2021-09-24
8271567: AArch64: AES Galois CounterMode (GCM) interleaved implementation using vector instructions

theRealAph:
This is a backport of the big AES/GCM patch from JDK head. It's a
major change and it's had very little time (almost a day) to mature in
head, so perhaps it shouldn't be backported to 11 for some time; I
wouldn't be at all surprised if some reviewers' reaction was "What
have you been smoking?" However, there is a good reason for a
backport: OpenJDK on x86 has a major advantage. AES/GCM is an
important cipher, the current AArch64 implementation is much slower
than x86, and some workloads are severely impacted.

I'm open to all arguments about why this should or shouldn't be pushed,
and I'm quite happy to wait for another release cycle or two if people
think that's the best course of action.

---------
### Progress

### Issue
 * [JDK-8271567](https://bugs.openjdk.java.net/browse/JDK-8271567): AArch64: AES Galois CounterMode (GCM) interleaved implementation using vector instructions

### Reviewers
 * [Xin Liu](https://openjdk.java.net/census#xliu) (@navyxliu - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/410/head:pull/410` \
`$ git checkout pull/410`

`$ git checkout pull/410` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/410/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 410`

`$ git pr show -t 410`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/410.diff">https://git.openjdk.java.net/jdk11u-dev/pull/410.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -309,7 +309,7 @@ void VM_Version::get_processor_features() {
       UseAES = true;
     }
     if (FLAG_IS_DEFAULT(UseAESCTRIntrinsics)) {
-      FLAG_SET_DEFAULT(UseAESCTRIntrinsics, true);
+      FLAG_SET_DEFAULT(UseAESCTRIntrinsics, false);
     }
   } else {
     if (UseAES) {
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -308,6 +308,9 @@ void VM_Version::get_processor_features() {
       warning("UseAESIntrinsics enabled, but UseAES not, enabling");
       UseAES = true;
     }
+    if (FLAG_IS_DEFAULT(UseAESCTRIntrinsics)) {
+      FLAG_SET_DEFAULT(UseAESCTRIntrinsics, true);
+    }
   } else {
     if (UseAES) {
       warning("AES instructions are not available on this CPU");
@@ -317,11 +320,10 @@ void VM_Version::get_processor_features() {
       warning("AES intrinsics are not available on this CPU");
       FLAG_SET_DEFAULT(UseAESIntrinsics, false);
     }
-  }
-
-  if (UseAESCTRIntrinsics) {
-    warning("AES/CTR intrinsics are not available on this CPU");
-    FLAG_SET_DEFAULT(UseAESCTRIntrinsics, false);
+    if (UseAESCTRIntrinsics) {
+      warning("AES/CTR intrinsics are not available on this CPU");
+      FLAG_SET_DEFAULT(UseAESCTRIntrinsics, false);
+    }
   }
 
   if (FLAG_IS_DEFAULT(UseCRC32Intrinsics)) {
--- a/src/hotspot/cpu/aarch64/globals_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/globals_aarch64.hpp
@@ -117,12 +117,10 @@ define_pd_global(intx, InlineSmallCode,          1000);
   product(intx, BlockZeroingLowLimit, 256,                              \
           "Minimum size in bytes when block zeroing will be used")      \
           range(1, max_jint)                                            \
-  product(bool, TraceTraps, false, "Trace all traps the signal handler") \
+  product(bool, TraceTraps, false, "Trace all traps the signal handler")\
   product(int, SoftwarePrefetchHintDistance, -1,                        \
           "Use prfm hint with specified distance in compiled code."     \
           "Value -1 means off.")                                        \
-          range(-1, 4096)                                               \
-  product(bool, UseAESGCMIntrinsics, false,                             \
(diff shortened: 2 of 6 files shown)


simonis:
I have no problem with you smoking whatsoever you like :)

But more serious. This isn't a full review but from a quick check it looks like this feature can easily turned off with `-XX:-UseAESCTRIntrinsics`, right? So in that case I have no objections of downporting this instantly to 11u. We might even be ultra-conservative and make `-XX:-UseAESCTRIntrinsics` the default in the downport, but I'm not sure it that's necessary.

Thanks for implementing this intrinsic and dowporting to 11u. It closes a real gap between x64 and aarch64,
Volker

theRealAph:
> But more serious. This isn't a full review but from a quick check it looks like this feature can easily turned off with `-XX:-UseAESCTRIntrinsics`, right? So in that case I have no objections of downporting this instantly to 11u. We might even be ultra-conservative and make `-XX:-UseAESCTRIntrinsics` the default in the downport, but I'm not sure it that's necessary.

I hadn't considered that idea, but it makes sense to turn it off by default. The problem is, I guess, that many customers won't know about it, would have poor performance, and maybe blame AArch64 servers.

shipilev:
Since this is math-heavy crypto code, I would rather wait for somebody to use/real-test the upstream implementation first. Unfortunately, timing is against us for JDK 18, as it would release only in March, and so if we want JDK 18 to be proven to work first, this would mean slipping the 11u backport to July 2022.

So I would propose this: wait another 6..8 weeks to see if AArch64 regressions are reported in mainline, then backport this to 11u for January 2022, disabled by default. Have a 11u-specific issue to remember enabling it by default after JDK 18 graduates and people get exposed to this code by default in JDK 18.

It would also help if we were able to throw some targeted testing for mainline code. Are there known good crypto test suites that could poke the holes in this implementation?

shipilev:
Now that I wrote this, another question: what about 17u? I think it is easier/cleaner to backport there first.

theRealAph:
> Since this is math-heavy crypto code, I would rather wait for somebody to use/real-test the upstream implementation first. Unfortunately, timing is against us for JDK 18, as it would release only in March, and so if we want JDK 18 to be proven to work first, this would mean slipping the 11u backport to July 2022.
> 
> So I would propose this: wait another 6..8 weeks to see if AArch64 regressions are reported in mainline, then backport this to 11u for January 2022, disabled by default.

OK. I'm not sure that disabling by default much reduces risk, because this patch does a fair bit of refactoring and commoning, even if the fast paths aren't enabled.

> Have a 11u-specific issue to remember enabling it by default after JDK 18 graduates and people get exposed to this code by default in JDK 18. This way we could also see some opt-in 11u testing (which is presumably a larger population than JDK 18 adopters).

Seems reasonable.

> It would also help if we were able to throw some targeted testing for mainline code. Are there known good crypto test suites that could poke the holes in this implementation?

Authenticated encryption is mostly self-testing: that's the idea.
I'm trying to think of where the errors might be. 

- Buffer overruns / other heap corruption.
- Incorrect encryption/decryption when update() is called with oddly-sized byte[].
- Failure to detect bad authentication tag.

The upper layers are passed the auth data, which they check, and the jtreg tests run the intrinsic against the pure-Java code. That should be enough to make sure we don't have an incorrect (but self-consistent) implementation.

External test suites might not run for long enough to test the intrinsic. I'll have a look at Wycheproof.

theRealAph:
I've reworked this to leave the existing code, so that unless `-XX:-UseAESCTRIntrinsics` is used on the command line this patch is entirely disabled.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -308,6 +308,9 @@ void VM_Version::get_processor_features() {
       warning("UseAESIntrinsics enabled, but UseAES not, enabling");
       UseAES = true;
     }
+    if (FLAG_IS_DEFAULT(UseAESCTRIntrinsics)) {
+      FLAG_SET_DEFAULT(UseAESCTRIntrinsics, false);
+    }
```
shipilev, line 313: `UseAESCTRIntrinsics` is already `false` by default, so this seems to be excessive. No harm doing this anyway.

shipilev (approved):
Looks fine to me.

RealCLanger:
@theRealAph, could you trigger a GHA run by first checking that it is enabled here - https://github.com/theRealAph/jdk11u-dev/actions - and then merging in current master branch?

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
navyxliu: +1

navyxliu (approved):
LGTM.  
I wish this is on by default, but I understand your concern. `UseAESCTRIntrinsics` is a diagnostic parameter. At least, it will allow bravehearts to try it out.

PR #486, opened 2021-10-06
Backport 6cbf57aa377c3e6edf79f31927e5cea69ef90a1d

GoeLin:
I backport this for parity with 11.0.14-oracle.

I had to resolve some chunks due to context differences.
Other places did not apply because there are only
test cases up to 10 in the current test.
I added the test cases New11 and New14, they seem 
to make sense for 11 now.
I had to undo the text blocks.

Unfortunately "8211148: var in implicit lambdas shouldn't be accepted
for source < 11" is not in 11, so New11 can be compiled with 
-source 10.  I commented out that test case. 
Test is passing this way.


RealCLanger:
> Unfortunately "8211148: var in implicit lambdas shouldn't be accepted for source < 11" is not in 11, so New11 can be compiled with -source 10. I commented out that test case. Test is passing this way.

I assume you'll update this one after you backported 8211148 with #497, right?

GoeLin:
"8211148: var in implicit lambdas shouldn't be accepted for source < 11" was pushed to jdk11u-dev, so I updated the test to also check the difference between 10 and 11.

TheRealMDoerr (approved):
New version LGTM.

PR #487, opened 2021-10-06
Backport 24bf35f862e285eeca662b9829901c0f91d247d5

GoeLin:
I want to downport this to jdk11u-dev so that downporting the follow up 
"JDK-8268714: [macos-aarch64] 7 java/net/httpclient/websocket tests failed"
is more simple.

I had to resolve the following parts:

Copyright in several files.

PlainHttpConnection.java  Change in a comment that was added in "8256459: java/net/httpclient/ManyRequests.java and java/net/httpclien..." does not apply. Skipped.

ProblemList: Tests removed were not problem listed.

PendingOperations.java: 
 * imports due to context changes
 * cleanup(): Changes from "8249786: java/net/httpclient/websocket/PendingPingTextClose.java fails very infrequently" are missing. I added the output to System.err from that change.

PendingPingTextClose.java: Different indentation and line breaking.

PendingPongTextClose.java: Different context.

### DIFF
--- a/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java
+++ b/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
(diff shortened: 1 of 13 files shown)


theRealAph:
From the discussion at https://github.com/openjdk/jdk17/pull/79, this is a dubious fix for some rather dubious tests. It touches a lot of shared runtime code which isn't broken. Surely this is a classic case of It Ain't Broke, Don't Fix It.

GoeLin:
OK, I drop this. See also my reasoning in the comments of JDK-8268714.
(Actually the test changes do not depend on the runtime changes. So they also could just be omitted.)

PR #488, opened 2021-10-06
8267652: c2 loop unrolling by 8 results in reading memory past array

VladimirKempik:
The backport can't be applied clean to jdk11 ( due to the miss of 8223347: Integration of Vector API (Incubator) )  and needed big rework.

The main idea behind the backport is to add new check to functions with VecS or VecD argument ( those who can read less than 16 bytes from memory) and not touch VecX/VecY/VecZ types.

The problematic place before the patch :

  0x000000011dd55603: vmovq  0x10(%r10,%rsi,1),%xmm0
  0x000000011dd5560a: vpxor  0x10(%r11,%rsi,1),%xmm0,%xmm0
  0x000000011dd55611: vmovq  %xmm0,0x10(%r13,%rsi,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
                                                ; - repro::xor_array@18 (line 12)

and after the patch:

  0x000000011f9a4d95: vmovq  0x10(%r11,%rsi,1),%xmm0
  0x000000011f9a4d9c: vmovq  0x10(%r10,%rsi,1),%xmm1
  0x000000011f9a4da3: vpxor  %xmm1,%xmm0,%xmm0
  0x000000011f9a4da7: vmovq  %xmm0,0x10(%r13,%rsi,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
                                                ; - repro::xor_array@18 (line 12)

Testing is pending.

### DIFF
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -1297,6 +1297,16 @@ class HandlerImpl {
 #endif
 };
 
+inline uint vector_length(const Node* n) {
+  const TypeVect* vt = n->bottom_type()->is_vect();
+  return vt->length();
+}
+
+inline uint vector_length_in_bytes(const Node* n) {
+  const TypeVect* vt = n->bottom_type()->is_vect();
+  return vt->length_in_bytes();
+}
+
 %} // end source_hpp
 
 source %{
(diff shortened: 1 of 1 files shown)


TheRealMDoerr:
Thanks for backporting it! Looks like you have addressed all required nodes. So, I think it's good. We'll test the fix.

PR #490, opened 2021-10-07
8237589: Fix copyright header formatting

RealCLanger:
This is a backport of 8237589. I applied the original change and had then to resolve several places where copyright changed in the meanwhile in 11u. Also, a few files that were touched in the original change don't exist in 11u.


RealCLanger:
I'd say this one is good. No regressions in SAP testing. JDK-8237601 does not apply since MaxDiagsRecompile was only introduced in JDK13 with https://bugs.openjdk.java.net/browse/JDK-8220632.

shipilev (approved):
All right, fine.

RealCLanger:
Thanks for the review. Got the approval.

/integrate

PR #580, opened 2021-11-02
8234930: Use MAP_JIT when allocating pages for code cache on macOS

VladimirKempik:
Please review this backport of JDK-8234930 to jdk11u-dev.
This is a prerequest for mac-aarch64
It's not applying clean due to the miss of these two patches
JDK-8253639: Change os::attempt_reserve_memory_at parameter order
JDK-8253650: Cleanup: remove alignment_hint parameter from os::reserve_memory
But they aren't critical and the patch can easily be modified

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed

### Integration blocker
&nbsp;⚠️ Dependency #578 must be integrated first

### Issue
 * [JDK-8234930](https://bugs.openjdk.java.net/browse/JDK-8234930): Use MAP_JIT when allocating pages for code cache on macOS

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/580/head:pull/580` \
`$ git checkout pull/580`

Update a local copy of the PR: \
`$ git checkout pull/580` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/580/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 580`

View PR using the GUI difftool: \
`$ git pr show -t 580`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk11u-dev/pull/580.diff">https://git.openjdk.java.net/jdk11u-dev/pull/580.diff</a>

</details>

PR #710, opened 2021-12-08
8234930: Use MAP_JIT when allocating pages for code cache on macOS

VladimirKempik:
Hello, here is alternative backport of 8234930 to jdk11u.
Very minimalistic change for platforms other than macos, just one unused argument.
We have used this patch in zulu11 for more than a year now and found no issues with it on macos and other platforms.

Technically it's mostly the same as previous (3patch) version but has no prerequests.

---------
### Progress

### Issue
 * [JDK-8234930](https://bugs.openjdk.java.net/browse/JDK-8234930): Use MAP_JIT when allocating pages for code cache on macOS

### Reviewers
 * [Thomas Stuefe](https://openjdk.java.net/census#stuefe) (@tstuefe - **Reviewer**) ⚠️ Review applies to dfef95e36a7db62a776eff2bc36960588bd0872a
 * [Xin Liu](https://openjdk.java.net/census#xliu) (@navyxliu - no project role) ⚠️ Review applies to dfef95e36a7db62a776eff2bc36960588bd0872a

### Contributors
 * Anton Kozlov `<akozlov@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/710/head:pull/710` \
`$ git checkout pull/710`

`$ git checkout pull/710` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/710/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 710`

`$ git pr show -t 710`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/710.diff">https://git.openjdk.java.net/jdk11u-dev/pull/710.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/bsd/os_bsd.cpp
+++ b/src/hotspot/os/bsd/os_bsd.cpp
@@ -2014,8 +2014,17 @@ bool os::pd_commit_memory(char* addr, size_t size, bool exec) {
     return true;
   }
 #elif defined(__APPLE__)
-  if (::mprotect(addr, size, prot) == 0) {
-    return true;
+  if (exec) {
+    // Do not replace MAP_JIT mappings, see JDK-8234930
+    if (::mprotect(addr, size, prot) == 0) {
+      return true;
+    }
+  } else {
+    uintptr_t res = (uintptr_t) ::mmap(addr, size, prot,
+                                       MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0);
+    if (res != (uintptr_t) MAP_FAILED) {
+      return true;
+    }
   }
 #else
   uintptr_t res = (uintptr_t) ::mmap(addr, size, prot,
--- a/src/hotspot/os/bsd/os_bsd.cpp
+++ b/src/hotspot/os/bsd/os_bsd.cpp
@@ -2183,15 +2183,18 @@ static int anon_munmap(char * addr, size_t size) {
   return ::munmap(addr, size) == 0;
 }
 
-MACOS_ONLY(char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
+#if defined(__APPLE__)
+char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
                             size_t alignment_hint,
                             bool executable) {
   return anon_mmap(requested_addr, bytes, (requested_addr != NULL), executable);
-})
-NOT_MACOS(char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
+}
+#else
+char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
                             size_t alignment_hint) {
   return anon_mmap(requested_addr, bytes, (requested_addr != NULL));
-})
+}
+#endif
 
 bool os::pd_release_memory(char* addr, size_t size) {
   return anon_munmap(addr, size);
--- a/src/hotspot/os/bsd/os_bsd.cpp
+++ b/src/hotspot/os/bsd/os_bsd.cpp
@@ -22,6 +22,16 @@
  *
  */
 
+/*
+ * On macOS MAP_JIT cannot be used in conjunction with MAP_FIXED when mapping
+ * a page for codecache. Therefore our traditional technique of doing commit
+ * and uncommit - replacing a mapping with another one at the same address
+ * range but swapped MAP_NORESERVE - does not work.
+ * The "exec" flag basically means "its code cache" and it should be used
+ * consistently for the same mapping (reserve-commit-uncommit etc)
+ * This affects pd_reserve_memory, pd_commit_memory, pd_uncommit_memory functions
+ */
+
 // no precompiled headers
 #include "jvm.h"
 #include "classfile/classLoader.hpp"
--- a/src/hotspot/os/bsd/os_bsd.cpp
+++ b/src/hotspot/os/bsd/os_bsd.cpp
@@ -2174,7 +2174,6 @@ NOT_MACOS(char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
   return anon_mmap(requested_addr, bytes, (requested_addr != NULL));
 })
 
-
 bool os::pd_release_memory(char* addr, size_t size) {
   return anon_munmap(addr, size);
 }
@@ -2355,7 +2354,7 @@ char* os::pd_attempt_reserve_memory_at(size_t bytes, char* requested_addr) {
 
   // Bsd mmap allows caller to pass an address as hint; give it a try first,
   // if kernel honors the hint then we can return immediately.
-  char * addr = anon_mmap(requested_addr, bytes, false/*fixed*/);
+  char * addr = anon_mmap(requested_addr, bytes, false);
   if (addr == requested_addr) {
     return requested_addr;
   }
--- a/src/hotspot/os/bsd/os_bsd.cpp
+++ b/src/hotspot/os/bsd/os_bsd.cpp
@@ -2007,9 +2007,13 @@ static void warn_fail_commit_memory(char* addr, size_t size, bool exec,
 //       problem.
 bool os::pd_commit_memory(char* addr, size_t size, bool exec) {
   int prot = exec ? PROT_READ|PROT_WRITE|PROT_EXEC : PROT_READ|PROT_WRITE;
-#ifdef __OpenBSD__
+#if defined(__OpenBSD__)
   // XXX: Work-around mmap/MAP_FIXED bug temporarily on OpenBSD
   Events::log(NULL, "Protecting memory [" INTPTR_FORMAT "," INTPTR_FORMAT "] with protection modes %x", p2i(addr), p2i(addr+size), prot);
+  if (::mprotect(addr, size, prot) == 0) {
+    return true;
+  }
+#elif defined(__APPLE__)
   if (::mprotect(addr, size, prot) == 0) {
     return true;
   }
@@ -2091,11 +2095,22 @@ char *os::scan_pages(char *start, char* end, page_info* page_expected, page_info
 }
 
 
-bool os::pd_uncommit_memory(char* addr, size_t size) {
-#ifdef __OpenBSD__
+bool os::pd_uncommit_memory(char* addr, size_t size, bool exec) {
+#if defined(__OpenBSD__)
   // XXX: Work-around mmap/MAP_FIXED bug temporarily on OpenBSD
   Events::log(NULL, "Protecting memory [" INTPTR_FORMAT "," INTPTR_FORMAT "] with PROT_NONE", p2i(addr), p2i(addr+size));
   return ::mprotect(addr, size, PROT_NONE) == 0;
+#elif defined(__APPLE__)
+  if (exec) {
+    if (::madvise(addr, size, MADV_FREE) != 0) {
+      return false;
+    }
+    return ::mprotect(addr, size, PROT_NONE) == 0;
+  } else {
+    uintptr_t res = (uintptr_t) ::mmap(addr, size, PROT_NONE,
+        MAP_PRIVATE|MAP_FIXED|MAP_NORESERVE|MAP_ANONYMOUS, -1, 0);
+    return res  != (uintptr_t) MAP_FAILED;
+  }
 #else
   uintptr_t res = (uintptr_t) ::mmap(addr, size, PROT_NONE,
                                      MAP_PRIVATE|MAP_FIXED|MAP_NORESERVE|MAP_ANONYMOUS, -1, 0);
@@ -2110,7 +2125,7 @@ bool os::pd_create_stack_guard_pages(char* addr, size_t size) {
 // If this is a growable mapping, remove the guard pages entirely by
 // munmap()ping them.  If not, just call uncommit_memory().
 bool os::remove_stack_guard_pages(char* addr, size_t size) {
-  return os::uncommit_memory(addr, size);
+  return os::uncommit_memory(addr, size, !ExecMem);
 }
 
 // If 'fixed' is true, anon_mmap() will attempt to reserve anonymous memory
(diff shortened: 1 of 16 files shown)


src/hotspot/os/bsd/os_bsd.cpp
```diff
@@ -2007,12 +2007,16 @@ static void warn_fail_commit_memory(char* addr, size_t size, bool exec,
...
 bool os::pd_commit_memory(char* addr, size_t size, bool exec) {
   int prot = exec ? PROT_READ|PROT_WRITE|PROT_EXEC : PROT_READ|PROT_WRITE;
-#ifdef __OpenBSD__
+#if defined(__OpenBSD__)
   // XXX: Work-around mmap/MAP_FIXED bug temporarily on OpenBSD
   Events::log(NULL, "Protecting memory [" INTPTR_FORMAT "," INTPTR_FORMAT "] with protection modes %x", p2i(addr), p2i(addr+size), prot);
   if (::mprotect(addr, size, prot) == 0) {
     return true;
   }
+#elif defined(__APPLE__)
```
tstuefe, line 2016: I'm confused here. We reserve with MAP_NORESERVE, then, on commit, just mprotect? Will that remove the MAP_NORESERVE flag?

Why not do mprotect only if exec=true (which we treat here synonymous for "is code cache area")?

src/hotspot/share/runtime/os.hpp
```diff
@@ -343,7 +345,7 @@ class os: AllStatic {
   static void   commit_memory_or_exit(char* addr, size_t size,
                                       size_t alignment_hint,
                                       bool executable, const char* mesg);
-  static bool   uncommit_memory(char* addr, size_t bytes);
+  static bool   uncommit_memory(char* addr, size_t bytes, bool exec);
```
tstuefe, line 348: Please make this default=false too, and revert the hunks where this is explicitly set to false

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2622,7 +2622,7 @@ void JavaThread::create_stack_guard_pages() {
   } else {
     log_warning(os, thread)("Attempt to protect stack guard pages failed ("
       PTR_FORMAT "-" PTR_FORMAT ").", p2i(low_addr), p2i(low_addr + len));
-    if (os::uncommit_memory((char *) low_addr, len)) {
+    if (os::uncommit_memory((char *) low_addr, len, !ExecMem)) {
```
tstuefe, line 2625: with default=false, this change is not needed

src/hotspot/share/runtime/os.hpp
```diff
@@ -326,7 +327,8 @@ class os: AllStatic {
 
   static int    vm_allocation_granularity();
   static char*  reserve_memory(size_t bytes, char* addr = 0,
-                               size_t alignment_hint = 0, int file_desc = -1);
+                               size_t alignment_hint = 0, int file_desc = -1,
+                               bool executable = false);
```
tstuefe, line 331: Please make this default=false too, and revert the hunks where this is explicitly set to false

tstuefe (changes requested):
Hi Vladimir,

this is already a lot better.

And slowly the whole story comes back to me. IIRC the problem is that MAP_JIT cannot be used in conjunction with MAP_FIXED, right? Therefore we cannot replace mappings? Therefore our traditional technique of doing commit and uncommit - replacing a mapping with another one at the same address range but swapped MAP_NORESERVE - does not work?

If this is true, could you please add a comment somewhere at the start of the file explaining this, and refer in both uncommit and commit to that? 

Additionally, also modify the associated JBS issue text and add it there too? Feel free to use the same text, I just don't want that knowledge to be gone. I remember Anton had a hard time explaining things to me last year, so it's not that obvious.

---

Please also add a comment to os_bsd.cpp to state that the "exec" flag basically means "its code cache" and that it should be used consistently for the same mapping (reserve-commit-uncommit etc). 

---

The change can be made a lot smaller still. Sure we will divert from upstream over time, but keeping patches small helps delay that.

1) Please make the exec parameter to commit() and uncommit() default false and remove all invocations wich use false or !ExecMem explicitly. That is the lion's share of calls.
2) Please, instead of adding "exec" to every version of pd_commit/uncommit, only add it for MacOS like this:

```
pd_uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
pd_commit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
```

and do the same with the two places where this parameter actually matters on invocation (os.cpp and VirtualSpace). 

---

Thanks, Thomas

VladimirKempik:
> 2. Please, instead of adding "exec" to every version of pd_commit/uncommit, only add it for MacOS like this:
> 
> ```
> pd_uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
> pd_commit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
> ```
> 
> and do the same with the two places where this parameter actually matters on invocation (os.cpp and VirtualSpace).
> 
> Thanks, Thomas

Hello Thomas, that plan doesn't seem to work

```
/Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/runtime/os.hpp:349:71: error: too many arguments provided to function-like macro invocation
  static bool   uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool executable = false));
                                                                      ^
/Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/utilities/macros.hpp:422:9: note: macro 'MACOS_ONLY' defined here
#define MACOS_ONLY(code) code
```

tstuefe:
> > 2. Please, instead of adding "exec" to every version of pd_commit/uncommit, only add it for MacOS like this:
> > 
> > ```
> > pd_uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
> > pd_commit_memory(char* addr, size_t bytes MACOS_ONLY(, bool exec));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and do the same with the two places where this parameter actually matters on invocation (os.cpp and VirtualSpace).
> > Thanks, Thomas
> 
> Hello Thomas, that plan doesn't seem to work
> 
> ```
> /Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/runtime/os.hpp:349:71: error: too many arguments provided to function-like macro invocation
>   static bool   uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool executable = false));
>                                                                       ^
> /Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/utilities/macros.hpp:422:9: note: macro 'MACOS_ONLY' defined here
> #define MACOS_ONLY(code) code
> ```
> 
> placing whole line under macro seems to work
> 
> ```
> MACOS_ONLY(static bool   pd_uncommit_memory(char* addr, size_t bytes,
>                                               bool executable = false);)
>   NOT_MACOS(static bool   pd_uncommit_memory(char* addr, size_t bytes);)
> ```

Hmm, weird, we use this syntax in many places. Okay, if this turns out to be too ugly, feel free to ignore my request (2) and keep the platform-dependent changes in all versions of pd_commit/uncommit. I leave this up to you.

Thanks, Thomas

VladimirKempik:
>mm, weird, we use this syntax in many places. Okay, if this turns out to be too ugly, feel free to ignore my request (2) and keep the platform-dependent changes in all versions of pd_commit/uncommit. I leave this up to you.
> 
> Thanks, Thomas

lets see how updated version plays, we are down to patching just 4 files

src/hotspot/os/bsd/os_bsd.cpp
```diff
@@ -2007,12 +2007,16 @@ static void warn_fail_commit_memory(char* addr, size_t size, bool exec,
 //       problem.
 bool os::pd_commit_memory(char* addr, size_t size, bool exec) {
   int prot = exec ? PROT_READ|PROT_WRITE|PROT_EXEC : PROT_READ|PROT_WRITE;
-#ifdef __OpenBSD__
+#if defined(__OpenBSD__)
```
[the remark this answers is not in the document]
VladimirKempik: this is needed to have #elif defined(__APPLE__) later

src/hotspot/os/bsd/os_bsd.cpp
VladimirKempik: that's leftovers,  I will make this section to be same as in upstream, like https://github.com/openjdk/jdk11u-dev/pull/580/files#diff-1f93205c2e57bee432f8fb7a0725ba1dfdbe5b901ac63010ea0b43922e34ac12

VladimirKempik: Leftovers, will do the same way as in upstream, like here https://github.com/openjdk/jdk11u-dev/pull/580/files#diff-1f93205c2e57bee432f8fb7a0725ba1dfdbe5b901ac63010ea0b43922e34ac12

VladimirKempik:
I think most of the issues you have found should be resolved by now

src/hotspot/os/bsd/os_bsd.cpp
```diff
@@ -2091,11 +2114,23 @@ char *os::scan_pages(char *start, char* end, page_info* page_expected, page_info
 }
 
 
-bool os::pd_uncommit_memory(char* addr, size_t size) {
-#ifdef __OpenBSD__
+MACOS_ONLY(bool os::pd_uncommit_memory(char* addr, size_t size, bool executable))
+NOT_MACOS(bool os::pd_uncommit_memory(char* addr, size_t size)) {
```
tstuefe, line 2118: Out of curiosity, did you try and build BSD? Does anyone? Or does it just quietly bitrot away?

At some point maybe the delta between BSD and MacOS are large enough to split the bases (and maybe retire BSD unless someone maintains it)

src/hotspot/os/bsd/os_bsd.cpp
```diff
@@ -2142,10 +2183,15 @@ static int anon_munmap(char * addr, size_t size) {
...
+MACOS_ONLY(char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
+                            size_t alignment_hint,
+                            bool executable) {
+  return anon_mmap(requested_addr, bytes, (requested_addr != NULL), executable);
+})
+NOT_MACOS(char* os::pd_reserve_memory(size_t bytes, char* requested_addr,
                             size_t alignment_hint) {
   return anon_mmap(requested_addr, bytes, (requested_addr != NULL));
-}
+})
```
tstuefe, line 2194: Nitpicking and a matter of taste, but for multiline diffs I find #ifdef __APPLE__ easier to read.

tstuefe (approved):
Hi Vladimir,

I am fine with this. I'll give this version a spin in our tests tonight.

The MACOS_ONLY issue is weird though since we use _ONLY macros in a similar way in a number of places. Ah well.

The rest of the remarks is plain nitpicking. Feel free to follow or ignore them.

Thanks for this, this is way easier to manage than a string of large patches.

Cheers, Thomas

src/hotspot/os/bsd/os_bsd.cpp
VladimirKempik: Not me, but some may build it for free/open bsd

navyxliu:
> /Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/runtime/os.hpp:349:71: error: too many arguments provided to function-like macro invocation
  static bool   uncommit_memory(char* addr, size_t bytes MACOS_ONLY(, bool executable = false));
                                                                      ^
/Users/vkempik/ws/openjdk/git/jdk11u-dev/src/hotspot/share/utilities/macros.hpp:422:9: note: macro 'MACOS_ONLY' defined here
#define MACOS_ONLY(code) code

This doesn't work because comma here is also meaningful for the preprocessor. 
You can replace it with another macro and it will postpone to expand.  macros.hpp defines 'COMMA' for this purpose. 

 > static bool   uncommit_memory(char* addr, size_t bytes MACOS_ONLY(COMMA bool executable = false));


It's still not elegant. it's up to you to choose one.

navyxliu (approved):
LGTM.

navyxliu:
btw, I am not a Reviewer. we still need other reviewers approve on this.

VladimirKempik:
I think the review from Thomas is enough, yours is optional, still nice to have,

tstuefe:
@VladimirKempik, SAP nightlies ran through with your patch on Windows x64, all Linux platforms and Mac intel; no problems attributable that I could see. Note that the tests did not include MacOS aarch64, since we don't have the machines yet. But I expect you tested that. We also don't test 32-bit, Solaris, or AIX. AIX is supposedly tested by IBM now since SAP stepped away from it.

> I think the review from Thomas is enough, yours is optional, still nice to have,

Personally, I'm fine with it, but I'm not aware of the rules here.

IMHO in maintenance releases we should at least have the same number of reviewers we require on Head. After all, we are much closer to production in update releases; once out this hits many customer systems immediately.

Maybe not for simple patches that get downported and had been tested already in newer releases. But at least for complex stuff or for patches we develop explicitly for older releases.

Ping to @GoeLin ...

VladimirKempik:
> @VladimirKempik, SAP nightlies ran through with your patch on Windows x64, all Linux platforms and Mac intel; no problems attributable that I could see. Note that the tests did not include MacOS aarch64, since we don't have the machines yet. But I expect you tested that. We also don't test 32-bit, Solaris, or AIX. AIX is supposedly tested by IBM now since SAP stepped away from it.

Not exactly, none can test mac-arm on openjdk11 as it's not present there yet. I have made a sanity check on zulu11, applying this patch to it. C2/C1 still works which means this patch doesn't break memory management for codecache on mac-arm.
Solaris is another issue, as we can't build solaris from openjdk11 repo here, only from zulu11 repo. ( due to harfbuzz issues on jdk11u-dev).
Since this patch equals to "nothing" on platforms other than macos, it should be safe for them.
> 
> > I think the review from Thomas is enough, yours is optional, still nice to have,
> 
> Personally, I'm fine with it, but I'm not aware of the rules here.
> 
> IMHO in maintenance releases we should at least have the same number of reviewers we require on Head. After all, we are much closer to production in update releases; once out this hits many customer systems immediately.
> 
> Maybe not for simple patches that get downported and had been tested already in newer releases. But at least for complex stuff or for patches we develop explicitly for older releases.
> 
> Ping to @GoeLin ...

tstuefe:
> > @VladimirKempik, SAP nightlies ran through with your patch on Windows x64, all Linux platforms and Mac intel; no problems attributable that I could see. Note that the tests did not include MacOS aarch64, since we don't have the machines yet. But I expect you tested that. We also don't test 32-bit, Solaris, or AIX. AIX is supposedly tested by IBM now since SAP stepped away from it.
> 
> Not exactly, none can test mac-arm on openjdk11 as it's not present there yet. I have made a sanity check on zulu11, applying this patch to it. C2/C1 still works which means this patch doesn't break memory management for codecache on mac-arm. Solaris is another issue, as we can't build solaris from openjdk11 repo here, only from zulu11 repo. ( due to harfbuzz issues on jdk11u-dev). Since this patch equals to "nothing" on platforms other than macos, it should be safe for them.

I agree. It is fine from my side. Thank you for testing, and for preparing a minimal patch.

GoeLin:
thanks for reworking this.  It looks much better now, more safe for all the other platforms.
My choice of macros would have been the COMMA approach, but like this it is 
good, too.

PR #80, opened 2021-06-30
8261297: NMT: Final report should use scale 1

tstuefe:
Hi all,

this pull request contains a backport of commit 1740de2a from the openjdk/jdk repository.

The commit being backported was authored by Thomas Stuefe on 11 Feb 2021 and was reviewed by Zhengyu Gu.

Patch applies cleanly.

Thanks!

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed

### Issue
 * [JDK-8261297](https://bugs.openjdk.java.net/browse/JDK-8261297): NMT: Final report should use scale 1

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/80/head:pull/80` \
`$ git checkout pull/80`

Update a local copy of the PR: \
`$ git checkout pull/80` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/80/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 80`

View PR using the GUI difftool: \
`$ git pr show -t 80`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk11u-dev/pull/80.diff">https://git.openjdk.java.net/jdk11u-dev/pull/80.diff</a>

</details>

PR #32, opened 2021-06-15
8265773: incorrect jdeps message "jdk8internals" to describe a removed JDK internal API

AlexanderScherbatiy:
This is a direct backport of the fix `JDK-8265773: incorrect jdeps message "jdk8internals" to describe a removed JDK internal API` to jdk11u-dev.
The patch applies cleanly.

The fix has been built with the latest jdk11u-dev sources on Ubuntu 18.04 and manually tested with reactor-core-3.4.5.jar  lib.

The jdeps output without the fix is `jdk8internals`:
```
build/linux-x86_64-normal-server-release/jdk/bin/jdeps -s reactor-core-3.4.5.jar 
reactor-core-3.4.5.jar -> jdk8internals
reactor-core-3.4.5.jar -> java.base
reactor-core-3.4.5.jar -> java.logging
reactor-core-3.4.5.jar -> not found
```

and with the fix is `JDK removed internal API`
```
build/linux-x86_64-normal-server-release/jdk/bin/jdeps -s reactor-core-3.4.5.jar 
reactor-core-3.4.5.jar -> JDK removed internal API
reactor-core-3.4.5.jar -> java.base
reactor-core-3.4.5.jar -> java.logging
reactor-core-3.4.5.jar -> not found
```

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed

### Issue
 * [JDK-8265773](https://bugs.openjdk.java.net/browse/JDK-8265773): incorrect jdeps message "jdk8internals" to describe a removed JDK internal API

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/32/head:pull/32` \
`$ git checkout pull/32`

Update a local copy of the PR: \
`$ git checkout pull/32` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/32/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 32`

View PR using the GUI difftool: \
`$ git pr show -t 32`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk11u-dev/pull/32.diff">https://git.openjdk.java.net/jdk11u-dev/pull/32.diff</a>

</details>

PR #16, opened 2021-06-04
8253899: Make IsClassUnloadingEnabled signature match specification

VladimirKempik:
Almost clean backport of 8253899 to jdk11u-dev. The testcase got different filename extension (c, not cpp), that had to be corrected.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8253899](https://bugs.openjdk.java.net/browse/JDK-8253899): Make IsClassUnloadingEnabled signature match specification

### Reviewers
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/16/head:pull/16` \
`$ git checkout pull/16`

Update a local copy of the PR: \
`$ git checkout pull/16` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/16/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 16`

View PR using the GUI difftool: \
`$ git pr show -t 16`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk11u-dev/pull/16.diff">https://git.openjdk.java.net/jdk11u-dev/pull/16.diff</a>

</details>

PR #715, opened 2021-12-13
Backport dbc9e4b50cdda35f5712deaf440c49f50b9edc96

VladimirKempik:
Initial version of JEP-391 backport to jdk11u.
Build system changes are mostly clean, except cds disabling part, it's copy&paste from aix part.
Things needing attention: os_bsd_aarch64.cpp and W^X transitions.
serviceability agent is mostly clean.
This passed GHA_tier1 testing.
Full regression testing is running now on Azul's infra, will report/update PR when done.
Sharing this PR slightly earlier to let other interested parties to run the tests too.
Example of cross-building on intel mac (needs Xcode12/13):
sh configure --with-boot-jdk=/Library/Java/JavaVirtualMachines/zulu-11.jdk/Contents/Home/ --with-build-jdk=/Library/Java/JavaVirtualMachines/zulu-11.jdk/Contents/Home/ --disable-warnings-as-errors --openjdk-target=aarch64-apple-darwin --with-extra-cflags='-arch arm64' --with-extra-ldflags='-arch arm64' --with-extra-cxxflags='-arch arm64'

### DIFF
--- a/src/hotspot/share/runtime/os.hpp
+++ b/src/hotspot/share/runtime/os.hpp
@@ -81,6 +81,11 @@ enum ThreadPriority {        // JLS 20.20.1-3
   CriticalPriority = 11      // Critical thread priority
 };
 
+enum WXMode {
+  WXWrite,
+  WXExec
+};
+
 // Executable parameter flag for os::commit_memory() and
 // os::commit_memory_or_exit().
 const bool ExecMem = true;
@@ -963,6 +968,11 @@ class os: AllStatic {
     Thread* _thread;
     bool _done;
   };
+
+#if defined(__APPLE__) && defined(AARCH64)
+  // Enables write or execute access to writeable and executable pages.
+  static void current_thread_enable_wx(WXMode mode);
+#endif // __APPLE__ && AARCH64
 
 #ifndef _WINDOWS
   // Suspend/resume support
--- a/make/autoconf/build-aux/config.guess
+++ b/make/autoconf/build-aux/config.guess
@@ -1,6 +1,7 @@
 #!/bin/sh
 #
-# Copyright (c) 2012, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2012, 2021, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2021, Azul Systems, Inc. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -103,6 +104,14 @@ if [ "x$OUT" = x ]; then
     if [ `uname -m` = loongarch64 ]; then
       OUT=loongarch64-unknown-linux-gnu
     fi
+  fi
+fi
+
+# Test and fix cpu on macos-aarch64, uname -p reports arm, buildsys expects aarch64
+echo $OUT | grep arm-apple-darwin > /dev/null 2> /dev/null
+if test $? = 0; then
+  if [ `uname -m` = arm64 ]; then
+    OUT=aarch64`echo $OUT | sed -e 's/[^-]*//'`
   fi
 fi
 
--- a/make/autoconf/flags.m4
+++ b/make/autoconf/flags.m4
@@ -111,19 +111,25 @@ AC_DEFUN([FLAGS_SETUP_MACOSX_VERSION],
 [
   # Additional macosx handling
   if test "x$OPENJDK_TARGET_OS" = xmacosx; then
+    # The expected format for <version> is either nn.n.n or nn.nn.nn. See
+    # /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/AvailabilityVersions.h
+
     # MACOSX_VERSION_MIN specifies the lowest version of Macosx that the built
     # binaries should be compatible with, even if compiled on a newer version
     # of the OS. It currently has a hard coded value. Setting this also limits
     # exposure to API changes in header files. Bumping this is likely to
     # require code changes to build.
-    MACOSX_VERSION_MIN=10.9.0
+    if test "x$OPENJDK_TARGET_CPU_ARCH" = xaarch64; then
+      MACOSX_VERSION_MIN=11.00.00
+    else
+      MACOSX_VERSION_MIN=10.9.0
+    fi
     MACOSX_VERSION_MIN_NODOTS=${MACOSX_VERSION_MIN//\./}
 
     AC_SUBST(MACOSX_VERSION_MIN)
 
     # Setting --with-macosx-version-max=<version> makes it an error to build or
-    # link to macosx APIs that are newer than the given OS version. The expected
-    # format for <version> is either nn.n.n or nn.nn.nn. See /usr/include/AvailabilityMacros.h.
+    # link to macosx APIs that are newer than the given OS version.
     AC_ARG_WITH([macosx-version-max], [AS_HELP_STRING([--with-macosx-version-max],
         [error on use of newer functionality. @<:@macosx@:>@])],
         [
--- a/make/autoconf/hotspot.m4
+++ b/make/autoconf/hotspot.m4
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2011, 2018, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2011, 2021, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -271,6 +271,14 @@ AC_DEFUN_ONCE([HOTSPOT_ENABLE_DISABLE_CDS],
     ENABLE_CDS="false"
     if test "x$enable_cds" = "xyes"; then
       AC_MSG_ERROR([CDS is currently not supported on AIX. Remove --enable-cds.])
+    fi
+  fi
+
+  # Disable CDS on macos-aarch64
+  if test "x$OPENJDK_TARGET_OS" = "xmacosx" && test "x$OPENJDK_TARGET_CPU" = "xaarch64"; then
+    ENABLE_CDS="false"
+    if test "x$enable_cds" = "xyes"; then
+      AC_MSG_ERROR([CDS is currently not supported on macOS/aarch64. Remove --enable-cds.])
     fi
   fi
 
--- a/make/common/NativeCompilation.gmk
+++ b/make/common/NativeCompilation.gmk
@@ -1213,7 +1213,7 @@ define SetupNativeCompilationBody
                 # This only works if the openjdk_codesign identity is present on the system. Let
                 # silently fail otherwise.
                 ifneq ($(CODESIGN), )
-		  $(CODESIGN) -s "$(MACOSX_CODESIGN_IDENTITY)" --timestamp --options runtime \
+		  $(CODESIGN) -f -s "$(MACOSX_CODESIGN_IDENTITY)" --timestamp --options runtime \
 		      --entitlements $$(call GetEntitlementsFile, $$@) $$@
                 endif
   endif
--- a/make/hotspot/gensrc/GensrcAdlc.gmk
+++ b/make/hotspot/gensrc/GensrcAdlc.gmk
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2013, 2018, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2013, 2021, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -99,6 +99,9 @@ ifeq ($(call check-jvm-feature, compiler2), true)
     ADLCFLAGS += -DAIX=1
   else ifeq ($(OPENJDK_TARGET_OS), macosx)
     ADLCFLAGS += -D_ALLBSD_SOURCE=1 -D_GNU_SOURCE=1
+    ifeq ($(HOTSPOT_TARGET_CPU_ARCH), aarch64)
+      ADLCFLAGS += -DR18_RESERVED
+    endif
   else ifeq ($(OPENJDK_TARGET_OS), windows)
     ifeq ($(call isTargetCpuBits, 64), true)
       ADLCFLAGS += -D_WIN64=1
--- a/make/lib/Awt2dLibraries.gmk
+++ b/make/lib/Awt2dLibraries.gmk
@@ -562,7 +562,7 @@ else
        maybe-uninitialized class-memaccess
   HARFBUZZ_DISABLED_WARNINGS_clang := unused-value incompatible-pointer-types \
        tautological-constant-out-of-range-compare int-to-pointer-cast \
-       undef missing-field-initializers
+       undef missing-field-initializers deprecated-declarations c++11-narrowing
   HARFBUZZ_DISABLED_WARNINGS_microsoft := 4267 4244 4090 4146 4334 4819 4101 4068 4805 4138
   HARFBUZZ_DISABLED_WARNINGS_C_solstudio := \
         E_INTEGER_OVERFLOW_DETECTED \
--- a/make/lib/Lib-jdk.hotspot.agent.gmk
+++ b/make/lib/Lib-jdk.hotspot.agent.gmk
@@ -36,7 +36,7 @@ else ifeq ($(OPENJDK_TARGET_OS), solaris)
   SA_LDFLAGS := -mt
 
 else ifeq ($(OPENJDK_TARGET_OS), macosx)
-  SA_CFLAGS := -Damd64 -D_GNU_SOURCE -mno-omit-leaf-frame-pointer \
+  SA_CFLAGS := -D_GNU_SOURCE -mno-omit-leaf-frame-pointer \
       -mstack-alignment=16 -fPIC
   LIBSA_EXTRA_SRC := $(SUPPORT_OUTPUTDIR)/gensrc/jdk.hotspot.agent
 else ifeq ($(OPENJDK_TARGET_OS), windows)
--- a/src/hotspot/cpu/aarch64/globalDefinitions_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/globalDefinitions_aarch64.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2015, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -53,7 +53,7 @@ const bool CCallingConventionRequiresIntsAsLongs = false;
 
 #define THREAD_LOCAL_POLL
 
-#if defined(_WIN64)
+#if defined(__APPLE__) || defined(_WIN64)
 #define R18_RESERVED
 #define R18_RESERVED_ONLY(code) code
 #define NOT_R18_RESERVED(code)
--- a/src/hotspot/cpu/aarch64/interpreterRT_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interpreterRT_aarch64.cpp
@@ -170,7 +170,7 @@ void InterpreterRuntime::SignatureHandlerGenerator::pass_object() {
     __ cbnz(temp(), L);
     __ mov(r0, zr);
     __ bind(L);
-    static_assert(sizeof(jobject) == wordSize, "");
+    STATIC_ASSERT(sizeof(jobject) == wordSize);
     __ str(r0, Address(to(), next_stack_offset(sizeof(jobject))));
   }
 }
--- a/src/hotspot/cpu/aarch64/interpreterRT_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interpreterRT_aarch64.cpp
@@ -1,6 +1,7 @@
 /*
- * Copyright (c) 2003, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2020, Red Hat Inc. All rights reserved.
+ * Copyright (c) 2021, Azul Systems, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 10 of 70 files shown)


VladimirKempik:
Testing revelaed the need to include https://bugs.openjdk.java.net/browse/JDK-8262896 into this PR as well.
However that fix has one small issue: 
template<int BType, typename JniType = decltype(BasicTypeToJni<BType>::jni_type)>
warning: default template arguments for a function template are a C++11 extension [-Wc++11-extensions]

VladimirKempik:
Together with #716 the hotspot can be built without warnings, some warning in splashscreen yet.

theRealAph:
This is an extremely difficult patch to review. I'll be soliciting other people's advice to help understand the risk versus reward ratio to bring this in to mainline.
One thing to consider is that, given the good performance of Rosetta, x86 binaries can be used. So, in effect this is only a performance patch for Mac/AArch64. On the other hand, Mac/AArch64 has been working well for some time on the AArch64-port/11u repo.

VladimirKempik:
Rosetta's JDK performance is subpar.
Also one need to consider rosetta's bugs which might interfere with jdk.
And last point - rosetta won't be there forever.

theRealAph:
> Rosetta's JDK performance is subpar. Also one need to consider rosetta's bugs which might interfere with jdk. And last point - rosetta won't be there forever.

Sure, all true, although last time I looked Rosetta was only 40% worse. Rosetta won't be there forever, but it might outlive JDK 11. We'll have to take everything into account, and review this very carefully.

Just for my information: if I apply this, will it be essentially identical to aarch64-port/jdk11u ?

VladimirKempik:
> > Rosetta's JDK performance is subpar. Also one need to consider rosetta's bugs which might interfere with jdk. And last point - rosetta won't be there forever.
> 
> Sure, all true, although last time I looked Rosetta was only 40% worse. Rosetta won't be there forever, but it might outlive JDK 11. We'll have to take everything into account, and review this very carefully.
> 
> Just for my information: if I apply this, will it be essentially identical to aarch64-port/jdk11u ?

It's an outdated version there, basically what was in zulu11 by that time.
But the jep-391 (for jdk17) had evolved before reaching the final version, W^X is different at least
This PR is partly based on that old patch, for os_bsd_aarch64.cpp, rest is a backport of the newer JEP-391 from upstream.

theRealAph:
> > Just for my information: if I apply this, will it be essentially identical to aarch64-port/jdk11u ?
> 
> It's an outdated version there, basically what was in zulu11 by that time. But the jep-391 (for jdk17) had evolved before reaching the final version, W^X is different at least This PR is partly based on that old patch, for os_bsd_aarch64.cpp, rest is a backport of the newer JEP-391 from upstream.

Ouch! That's terrible, I had no idea. So everybody who is building AArch64 from non-Azul trees has a broken release.

VladimirKempik:
I was slightly wrong, it's an early version of jep-391 for jdk17, not for jdk11
We have never shared jdk11 backport on aarch64-port. We only shared it with JetBrainsRuntime11.

VladimirKempik:
About the tests:
Regression testing seems to be fine on macos_aarch64 and linux_aarch64
intel platform result is yet to be analyzed.
as for the failure in GHA_tier1_linux32_hotspot_runtime, it's strange, it wasn't failing for me ( in my branch) before I created this PR - https://github.com/VladimirKempik/jdk11u-dev/runs/4510727420?check_suite_focus=true
And I can't reproduce that issue locally.

VladimirKempik:
@theRealAph 
FYI, about possible consequences for other platforms in jdk11u-dev
JDK-8253816: Support macOS W^X
it's all under MACOS_AARCH64_ONLY, so isolated to just one platform

JDK-8253817: Support macOS Aarch64 ABI in Interpreter 
JDK-8253818: Support macOS Aarch64 ABI for compiled wrappers
Applied mostly clean from upstream ( where it was tested already), thanks to prerequests I have integrated.
In worst case could impact only other aarch64 platforms

JDK-8253819: Implement os/cpu for macOS/AArch64
this is purely about bsd_aarch64, not dangerous for any other platform

JDK-8253839: Update tests and JDK code for macOS/Aarch64
isolated to macos_aarch64

JDK-8254941: Implement Serviceability Agent for macOS/AArch64
this is mostly about adding the missing case for bsd_aarch64

JDK-8255776: Change build system for macOS/AArch64
needs to review that carefully, but it's mostly clean backport from jdk17, mostly about different filenames.

JDK-8262903: [macos_aarch64] Thread::current() called on detached thread
Sadly I don't remember what it was about, probably one macos_aarch64_only addtition to generis os_bsd code.

JDK-8262896: [macos_aarch64] Crash in jni_fast_GetLongField
this might affect other aarch64 platforms. it was tested in jdk17 already and applied mostly clean

So if you look at the patch and ignore everything about w^x and bsd_aarch64 then there is not very much left

VladimirKempik:
Tier1/Tier2 on macos-aarch64 is good, only one test failed from tier2: hotspot/jtreg/runtime/Metaspace/FragmentMetaspace.java
it should be fixed after #710 is integrated

phohensee:
Here's a Rosetta issue against Corretto: https://github.com/corretto/corretto-11/issues/183

theRealAph:
> @theRealAph FYI, about possible consequences for other platforms in jdk11u-dev JDK-8253816: Support macOS W^X it's all under MACOS_AARCH64_ONLY, so isolated to just one platform

Sure, but this is a failure of process. It means that AArch64/MacOS would be going straight in to 11u mainline without being first baked in AArch64-port/11u.

phohensee:
The bug is closed because the failure only happens when running under Rosetta, so it's likely a Rosetta issue. Also, we don't yet support Corretto on M1.

VladimirKempik:
More testing on intel platforms ( except solaris_intel) showed no regressions.

VladimirKempik:
@theRealAph @GoeLin 
So what would be the verdict, should we proceed with this review here or go first into the aarch64-port and stay there for a couple of more monthes ?

theRealAph:
> @theRealAph @GoeLin So what would be the verdict, should we proceed with this review here or go first into the aarch64-port and stay there for a couple of more monthes ?

I've been thinking a lot about this. At the minimum, the missing parts should go into a staging repo in aarch64-port, if only so that we have a record of what happened. I hope that it won't be a couple of months before it all goes in to mainline.

VladimirKempik:
> > @theRealAph @GoeLin So what would be the verdict, should we proceed with this review here or go first into the aarch64-port and stay there for a couple of more monthes ?
> 
> I've been thinking a lot about this. At the minimum, the missing parts should go into a staging repo in aarch64-port, if only so that we have a record of what happened. I hope that it won't be a couple of months before it all goes in to mainline.

Ok, then, can I ask you as a project lead of aarch64-port to create a branch (in https://github.com/openjdk/aarch64-port ) which will be a fork of current jdk11u-dev state ?

Then I will proceed with the same pull-request into aarch64-port project.

phohensee:
Andrew, would it be ok to do the backports one at a time in jdk11u-dev?

VladimirKempik:
> Andrew, would it be ok to do the backports one at a time in jdk11u-dev?

Excuse me, what do you mean ? all the prerequests are already in jdk11u-dev. This backport should go as a whole, it's not possible to split it in parts, parts alone won't work.

lewurm:
> Ok, then, can I ask you as a project lead of aarch64-port to create a branch (in https://github.com/openjdk/aarch64-port ) which will be a fork of current jdk11u-dev state ?

I created a branch based on the current state of jdk11u-dev here: https://github.com/openjdk/aarch64-port/tree/jdk11-macos

phohensee:
That's what I wanted to know. Thanks.

VladimirKempik:
> > Ok, then, can I ask you as a project lead of aarch64-port to create a branch (in https://github.com/openjdk/aarch64-port ) which will be a fork of current jdk11u-dev state ?
> 
> I created a branch based on the current state of jdk11u-dev here: https://github.com/openjdk/aarch64-port/tree/jdk11-macos

Thank you.

I have created a pull request now https://github.com/openjdk/aarch64-port/pull/14

theRealAph:
One meta-question: much of this affects os_cpu/bsd_aarch64, but we're testing on Apple. Is anyone testing on (e.g.) FreeBSD/arm64? If not, maybe we could reach out to the FreeBSD devs.

src/hotspot/share/runtime/os.hpp
```diff
@@ -81,6 +81,11 @@ enum ThreadPriority {        // JLS 20.20.1-3
   CriticalPriority = 11      // Critical thread priority
 };
 
+enum WXMode {
+  WXWrite,
+  WXExec
+};
+
```
adinn, line 88: Why is this defined for all arches rather than just for MACOS_AARCH64?

Is it ever meant to be used outside of a MACOS_AARCH64_ONLY macro or #ifdef

adinn (approved):
I think this is ok to approve. Most changes are isolated to MacOS/bsd && AArch64. Those that affect other archs appear to be making changes that preserve current behaviour.

The only thing I found questionable is defining enum WXMode for all architectures rather than just . However, since that managed to work its way in upstream without causing any issues I guess the question is somewhat academic.

theRealAph:
> I think this is ok to approve. Most changes are isolated to MacOS/bsd && AArch64. Those that affect other archs appear to be making changes that preserve current behaviour.
> 
> The only thing I found questionable is defining enum WXMode for all architectures rather than just . However, since that managed to work its way in upstream without causing any issues I guess the question is somewhat academic.

Yes, I don't think it hurts so leave it.
I have discussed this patch with FreeBSD maintainers, and they are maintaining their own patches. While I regret that this is necessary, I do understand. So, there is no reason to hold back this patch.

lewurm (approved):

ahmedmuhsin:
Backporting [8241004: NMT tests fail on unaligned thread size with debug  #build](https://github.com/openjdk/jdk/commit/0b20eafb30cc1de372141f2a427d13fdc6565e3a) fixes the second issue from my previous post.

VladimirKempik:
May you have some issues with jspawnhelper ?
an output of this command might help to understand more
codesign -dvvv lib/jspawnhelper

tstuefe:
Hi, 

I was asked off-list to have a quick look at these problems.

-----

About the runtime-exec related crashes. I looked a the hs-err file [forkAndExec_coredump.log](https://github.com/openjdk/jdk11u-dev/files/7927720/forkAndExec_coredump.log):

- We crash with a SIGBUS because of accessing an unaligned address: `siginfo: si_signo: 10 (SIGBUS), si_code: 1 (BUS_ADRALN), si_addr: 0x000000013000bbfc`
- because we try to execute code from there `pc=0x000000013000bbfc` and its not 64-bit aligned.
- Crash address points into the code heap, into interpreter stub for calling native methods: `method entry point (kind = native)  [0x000000013000b940, 0x000000013000c100]  1984 bytes`
- the native method involved is `java.lang.ProcessImpl.forkAndExec(I[B[B[BI[BI[B[IZ)I+0` but I assume we never entered it. Also see note that the crash offset is 0.

Atm I assume something is wrong with the native call stub generated by the template interpreter. Its weird though that this would manifest only when calling into this particular native method.

I don't think jspawnhelper is involved. Beside what I wrote above, we do crash in the parent process, before exec'ing jspawnhelper. I think we never even got around to fork.

It's weird that this is only happening in 11. Is it? 

Just for completeness I compared the native process handling code between 11 and 19, and see not much difference. That code seldom changed, and nothing relevant changed for MacOS. Most changes had to do with Solaris platform removal. Note that MacOS always used posix_spawn.

Looking at the code generating method entry stubs on aarch64, there are some differences between JDK19 and JDK11 in `TemplateInterpreterGenerator::generate_native_entry`. Could 11 miss some fixes here? E.g. there is mentioning of a race fixed in the context of JEP 376 (https://github.com/openjdk/jdk/blob/2ea0edf2c40edde4c191864a40e7a4d741ac0b8e/src/hotspot/cpu/aarch64/templateInterpreterGenerator_aarch64.cpp#L1333-L1340). Not sure how relevant that would be in older releases and outside ZGC though. 

I am not an aarch64 expert, maybe someone from RedHat could look at this (@theRealAph)  ?

-----

NMT crashes

I looked at https://github.com/openjdk/jdk11u-dev/files/7927880/assertCrash_coredump.log.

All NMT tests crash when trying to generate an NMT report via jcmd. They crash when walking stored information about virtual memory segments prior to printing, because there boundaries are not page aligned.

What confused me was that I thought MacOS m1 uses 64k pages. I tried to confirm this by looking at the crash dump (I wish we would just print out os::vm_page_size) but I found the point where we protect stack guards:

```
Event: 8.927 Protecting memory [0x00000001722b8000,0x00000001722c4000] with protection modes 0
```

and these boundaries are not 64-k page aligned. Could someone with an m1 Mac confirm that os::vm_page_size is 64k? Is there anything special on MacOS m1 wrt to page size? 

One possible reason would be that NMT also accounts for thread stacks too, and traditionally accounted those as if they were allocated using mmap. But there is nothing that says that posix threads need to align their stacks to page size. E.g. on AIX it was not so. So they may not be page-aligned, which would generate the kind of errors you see. 

Zhengyu fixed this code up in JDK 13 with https://bugs.openjdk.java.net/browse/JDK-8204552. Maybe a similar solution is needed for MacOS m1?

To quickly confirm or exclude this theory, you could just disable the accounting of thread stacks (e.g. by stubbing out `MemTracker::record_thread_stack` and `MemTracker::release_thread_stack`. Other than that, maybe @zhengyu123 can take a look too .

---------

Hope I could help,

Cheers, Thomas

theRealAph:
> Hi,
> 
> I was asked off-list to have a quick look at these problems.

Is any of this reproducible on MacOS 12?

tstuefe:
> > Hi,
> > I was asked off-list to have a quick look at these problems.
> 
> Is any of this reproducible on MacOS 12?

No idea, I don't have a Mac. Just looked at the crash logs.

theRealAph:
> Backporting [8241004: NMT tests fail on unaligned thread size with debug #build](https://github.com/openjdk/jdk/commit/0b20eafb30cc1de372141f2a427d13fdc6565e3a) fixes the second issue from my previous post.

OK, so that's just a backport we're missing.

tstuefe:
> > Backporting [8241004: NMT tests fail on unaligned thread size with debug #build](https://github.com/openjdk/jdk/commit/0b20eafb30cc1de372141f2a427d13fdc6565e3a) fixes the second issue from my previous post.
> 
> OK, so that's just a backport we're missing.

Ah, I missed that, sorry.

ahmedmuhsin:
> Is any of this reproducible on MacOS 12?

The forkAndExec crash did not occur on MacOS 12. The NMT issues did, but the missing backport fixed the issue for all OS versions.

VladimirKempik:
Thanks for analysis Thomas.
At Azul we have a testsuite to check java<->jni_native arguments passing on lots of different arguments combinations.
I have run them with fastdebug build on both macos11 and macos12 and found no failures

>It's weird that this is only happening in 11. Is it?
It's very weird.

The issue is intermittent which make me think it could be a rare corner case in interp wrapper generator.

If it were one more R18 failure then it would likely crash on macos12 and not macos11

ahmedmuhsin:
> @ahmedmuhsin, could you possible try to replicate the issue with zulu11 ?
> https://www.azul.com/downloads/?version=java-11-lts&os=macos&architecture=arm-64-bit&package=jdk
> also, you test very old version of macos11.0.1 (Darwin 20.1.0)
> IIRC we saw very strange issues (not even java related) with first versions of macos_arm too, would it be possible for you to recheck with macos 11.4+ ?

@VladimirKempik  I'll give it a shot. 
Thanks for the input everyone.

tstuefe:
> btw, https://developer.apple.com/documentation/xcode/investigating-memory-access-crashes
> 
> EXC_ARM_DA_ALIGN. The crashed thread tried to access memory that isn’t appropriately aligned. This exception code is rare because 64-bit ARM CPUs work with misaligned data. However, you may see this exception subtype if **the memory address is both misaligned and located in an unmapped memory region**. You may have other crash reports that show a memory access issue with a different exception subtype, which are likely caused by the same underlying memory access issue.

Is this also true for code? I would have thought that instructions have to be loaded from a 64-bit-aligned address?

theRealAph:
> Is this also true for code? I would have thought that instructions have to be loaded from a 64-bit-aligned address?

It's 32 bits: 4 bytes.

ahmedmuhsin:
@VladimirKempik, Apologies about the delay :)

I have been busy with other tasks since my last post. Nevertheless, today I was able to focus on testing the issue based on your suggestions. I tested  both my build of this branch and the zulu11 build on MacOS versions `11.0`, `11.2.1`, `11.5.2`, and `11.6.2`.  Here are my results:

| OS Version |  My Build  |  Zulu11 |
| :---        |    :----:   |          ---: |
| 11.0 | Crash | No Crash |
| 11.2.1 | Crash | No Crash |
| 11.5.2 | No Crash | No Crash |
| 11.6.2 | No Crash | No Crash |

VladimirKempik:
Great, thanks, I have access to some 11.1 host and will try to replicate the issue there.

VladimirKempik:
Crashlog with hsdis - https://pastebin.com/5kz2vfEH

#  SIGBUS (0xa) at pc=0x000000012100bbfc, pid=76024, tid=39683
siginfo: si_signo: 10 (SIGBUS), si_code: 1 (BUS_ADRALN), si_addr: 0x000000012100bbfc

`	// #128
  0x000000012100bbc0: movk	x8, #0x2100, lsl #16
  0x000000012100bbc4: movk	x8, #0x1, lsl #32
  0x000000012100bbc8: br	x8
  0x000000012100bbcc: ldp	x22, x24, [x29, #-72]
  0x000000012100bbd0: ldur	x12, [x29, #-24]
  0x000000012100bbd4: ldr	x10, [x12, #80]
  0x000000012100bbd8: add	x0, x28, #0x350
  0x000000012100bbdc: adr	x8, 0x000000012100bbfc
  0x000000012100bbe0: str	x8, [x28, #824]
  0x000000012100bbe4: str	x20, [x28, #816]
  0x000000012100bbe8: str	x29, [x28, #832]
  0x000000012100bbec: orr	x8, xzr, #0x4
  0x000000012100bbf0: add	x9, x28, #0x3c8
  0x000000012100bbf4: stlr	w8, [x9]
  0x000000012100bbf8: blr	x10
  0x000000012100bbfc: isb
  0x000000012100bc00: ldur	x12, [x29, #-24]
  0x000000012100bc04: sub	x8, x20, #0x20
  0x000000012100bc08: and	sp, x8, #0xfffffffffffffff0
  0x000000012100bc0c: str	d0, [x20, #-16]!
  0x000000012100bc10: str	xzr, [x20, #-8]!
  0x000000012100bc14: str	x0, [x20, #-8]!
  0x000000012100bc18: mov	x8, #0x5  `

VladimirKempik:
Looks like it was two missing WX transitions in JNI_QUICK_ENTRY and JVM_QUICK_ENTRY
These transitions are present in zulu11
JNI_QUICK_ENTRY and JVM_QUICK_ENTRY macros are missing in jdk17/upstream.

That explains everything, except the macos version reproducibility.

@ahmedmuhsin, when you will have a time, please check the updated PR on macos 11.0.1 or 11.2.1

ahmedmuhsin:
@VladimirKempik I tested again on all the OS versions that I mentioned above and everything looks very good. No crashes. Thanks!

VladimirKempik:
@rnkovacs , parts of your jep-388 backport reverted ( https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1495, https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1558, https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1671 ) results in more warning for this PR:

/Users/runner/work/jdk11u-dev/jdk11u-dev/jdk/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp:1495:52: error: format specifies type 'unsigned long long' but the argument has type 'uintptr_t' (aka 'unsigned long') [-Werror,-Wformat]
    snprintf(buffer, sizeof(buffer), PTR64_FORMAT, imm64);
                                     ~~~~~~~~~~~~  ^~~~~
1 error generated.

was these three changes really needed or can I revert them ?

theRealAph:
On 2/8/22 06:48, VladimirKempik wrote:
> @rnkovacs , parts of your jep-388 backport reverted ( https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1495, https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1558, https://github.com/VladimirKempik/jdk11u-dev/commit/7bdb8ac7f9f9#diff-0f4150a9c607ccd590bf256daa800c0276144682a92bc6bdced5e8bc1bb81f3aL1671 ) results in more warning for this PR:
> 
> /Users/runner/work/jdk11u-dev/jdk11u-dev/jdk/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp:1495:52: error: format specifies type 'unsigned long long' but the argument has type 'uintptr_t' (aka 'unsigned long') [-Werror,-Wformat]
>      snprintf(buffer, sizeof(buffer), PTR64_FORMAT, imm64);
>                                       ~~~~~~~~~~~~  ^~~~~
> 1 error generated.
> 
> was these three changes really needed or can I revert them ?

What does this refer to? I can't find any such thing in the current state of the patch.

VladimirKempik:
These three line changes are part of jep-388 (win-arm64) backport to jdk11u-dev, they cause troubles ( warnings which are under Werror becomes an issue)
these changes aren't present in upstream (jdk19), for example - https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp#L1316

so I wonder why did they became part of jep-388 backport to jdk11u-dev
if there is good reason, then I would like to revert these changes as part of this PR.

theRealAph:
> so I wonder why did they became part of jep-388 backport to jdk11u-dev if there is no good reason, then I would like to revert these changes as part of this PR.

There can be no good reason for an incorrect change, so please do that.

VladimirKempik:
Ok.
Also. After latest merge with master this can now be built without -with-extra-"c/ld/cxx"flags

ahmedmuhsin:
@VladimirKempik should I submit a PR to backport [8241004: NMT tests fail on unaligned thread size with debug build](https://github.com/openjdk/jdk/commit/0b20eafb30cc1de372141f2a427d13fdc6565e3a) or is it already on your radar?

The backport solves issue 2 from my post here: https://github.com/openjdk/jdk11u-dev/pull/715#issuecomment-1020401252

VladimirKempik:
Please proceed with 8241004, I have few other macarm related issues on my radar atm.

VladimirKempik:
@theRealAph what do you think, is this PR ready to go jdk11u-dev ?
it would be good to merge it some time before rdp2 to have a time for non-critical fixes.

theRealAph (approved):
This looks good enough.

VladimirKempik:
Oracle has pulled this into their java11 - https://bugs.openjdk.java.net/browse/JDK-8275573

@GoeLin is there anything holding us from getting approval for 11u ?

PR #771, opened 2022-01-17
8236210: javac generates wrong annotation for fields generated from record components

cushon:
This is a backport of [JDK-8236210: javac generates wrong annotation for fields generated from record components](https://bugs.openjdk.java.net/browse/JDK-8236210)

This is a partial backport.

* The focus of JDK-8236210 was a bug with records and most of the diff was to logic that was added to deal with records. However the patch also included a fix to the logic for checking annotation applicability in `Check.annotationApplicable` that fixes a bug with annotations on `var`, so I backported the part that is relevant to Java 11.
* I added a test for the issue specific to `var`, since the original test is for records and cannot be backported.

Testing: x86 build, affected tests, tier1

---------
### Progress

### Issue
 * [JDK-8236210](https://bugs.openjdk.java.net/browse/JDK-8236210): javac generates wrong annotation for fields generated from record components

### Reviewers
 * [Paul Hohensee](https://openjdk.java.net/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/771/head:pull/771` \
`$ git checkout pull/771`

`$ git checkout pull/771` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/771/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 771`

`$ git pr show -t 771`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/771.diff">https://git.openjdk.java.net/jdk11u-dev/pull/771.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
@@ -3084,8 +3084,19 @@ public class Check {
 
     /** Is the annotation applicable to the symbol? */
     boolean annotationApplicable(JCAnnotation a, Symbol s) {
+        Optional<Set<Name>> targets = getApplicableTargets(a, s);
+        /* the optional could be emtpy if the annotation is unknown in that case
+         * we return that it is applicable and if it is erroneous that should imply
+         * an error at the declaration site
+         */
+        return targets.isEmpty() || targets.isPresent() && !targets.get().isEmpty();
+    }
+
+    @SuppressWarnings("preview")
+    Optional<Set<Name>> getApplicableTargets(JCAnnotation a, Symbol s) {
         Attribute.Array arr = getAttributeTargetAttribute(a.annotationType.type.tsym);
         Name[] targets;
+        Set<Name> applicableTargets = new HashSet<>();
 
         if (arr == null) {
             targets = defaultTargetMetaInfo(a, s);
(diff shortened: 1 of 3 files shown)


phohensee (approved):
Lgtm.

GoeLin:
Hi Liam,
your change does not build with Java 10:
src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java:3092: error: cannot find symbol
        return targets.isEmpty() || targets.isPresent() && !targets.get().isEmpty();
                      ^
  symbol:   method isEmpty()
  location: variable targets of type Optional<Set<Name>>

See also DEFAULT_ACCEPTABLE_BOOT_VERSIONS in make/autoconf/version-numbers

cushon:
@GoeLin thanks for the catch! I did my testing with 11 as the bootstrap JDK instead of with 10. I have pushed a fix to remove the use of `Optional.isEmpty`.

phohensee (approved):
Lgtm again.

GoeLin:
Thanks for fixing. Maybe we should restrict us to only 11 for bootjdk at some point.

jerboaa:
> Maybe we should restrict us to only 11 for bootjdk at some point.

That wouldn't be a good move as new bootstrapping work would need to first bootstrap some special version of JDK 11 (perhaps GA?) with JDK 10 and then do it again for the current version of JDK 11 with JDK 11 itself.

PR #784, opened 2022-01-20
Backport 7c6f57fcb1f1fcecf26f7b8046a5a41ca6d9c315

GoeLin:
I backport this for parity with 11.0.15-oracle.

I had to adapt the test.
@key randomness is not supported in 11, as are -XX:StressSeed=42 and -XX:+StressIGVN.
I removed these.  This makes the second @run call superfluous.

Still, the test crashes the VM without the fix.


phohensee (approved):
Lgtm.

PR #805, opened 2022-02-03
8255410: Add ChaCha20 and Poly1305 support to SunPKCS11 provider

GoeLin:
I backport this for parity with 11.0.15-oracle.

I had to do a row of adaptions. There are some implementation
differences, and many Java 17 usages that had to be changed.

In the original change in SunPKCS11.java, dA() is used 
to give alias names of algorithms. It calls to 
SecurityProviderConstants.getAliases().
In SunPKCS11.java of jdk11 neither dA() nore getAliases()
are available. It uses d() instead of da() and calls a 
s() with a list of literal strings to give the alias names.
SecurityProviderConstants.java is in java.base. The jdk11 
version does not contain the list of aliases that can be 
found in 17.

I looked up whether there are aliases listed for 
"ChaCha20-Poly1305" in 17, but found none, so I added
an empty call s() in SunPKCS11.java.

In P11AEADCipher.java I had to modify the syntax of 
a switch statement.

In CK_SALSA20_CHACHA20_POLY1305_PARAMS.java
I had to replace HexFormat.of().formatHex(...)
which is a java.util class not in 11.

The tests use HexFormat heavily.
I replaced it by HexToBytes() from TestKATForGCM and 
by toHexString() from TestLeadingZeroesP11.java.
To make these methods available I moved them to the 
superclass PKCS11Test.java. 
This way they can be used for potential later backports, too.
I also had to adapt a switch statement using '->' to syntax
know to 11.

---------
### Progress
- [x] Change requires a CSR request to be approved

### Issues
 * [JDK-8255410](https://bugs.openjdk.java.net/browse/JDK-8255410): Add ChaCha20 and Poly1305 support to SunPKCS11 provider
 * [JDK-8265008](https://bugs.openjdk.java.net/browse/JDK-8265008): Add ChaCha20 and Poly1305 support to SunPKCS11 provider (**CSR**)

### Reviewers
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/805/head:pull/805` \
`$ git checkout pull/805`

`$ git checkout pull/805` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/805/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 805`

`$ git pr show -t 805`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/805.diff">https://git.openjdk.java.net/jdk11u-dev/pull/805.diff</a>

</details>

### DIFF
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -39,6 +39,8 @@ import javax.security.auth.login.FailedLoginException;
 import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.PasswordCallback;
+
+import com.sun.crypto.provider.ChaCha20Poly1305Parameters;
 
 import jdk.internal.misc.InnocuousThread;
 import sun.security.util.Debug;
@@ -600,6 +602,8 @@ public final class SunPKCS11 extends AuthProvider {
                 m(CKM_AES_KEY_GEN));
         d(KG,  "Blowfish",      P11KeyGenerator,
                 m(CKM_BLOWFISH_KEY_GEN));
+        d(KG,  "ChaCha20",      P11KeyGenerator,
+                m(CKM_CHACHA20_KEY_GEN));
 
         // register (Secret)KeyFactories if there are any mechanisms
         // for a particular algorithm that we support
@@ -626,6 +630,11 @@ public final class SunPKCS11 extends AuthProvider {
         d(AGP, "GCM",            "sun.security.util.GCMParameters",
                 m(CKM_AES_GCM));
 
+        d(AGP, "ChaCha20-Poly1305",
+                "com.sun.crypto.provider.ChaCha20Poly1305Parameters",
+                s(),
+                m(CKM_CHACHA20_POLY1305));
+
         d(KA, "DH",             P11KeyAgreement,        s("DiffieHellman"),
                 m(CKM_DH_PKCS_DERIVE));
         d(KA, "ECDH",           "sun.security.pkcs11.P11ECDHKeyAgreement",
@@ -642,6 +651,8 @@ public final class SunPKCS11 extends AuthProvider {
                 m(CKM_AES_CBC));
         d(SKF, "Blowfish",      P11SecretKeyFactory,
                 m(CKM_BLOWFISH_CBC));
+        d(SKF, "ChaCha20",      P11SecretKeyFactory,
+                m(CKM_CHACHA20_POLY1305));
 
         // XXX attributes for Ciphers (supported modes, padding)
         d(CIP, "ARCFOUR",                       P11Cipher,      s("RC4"),
@@ -708,6 +719,11 @@ public final class SunPKCS11 extends AuthProvider {
                 m(CKM_BLOWFISH_CBC));
         d(CIP, "Blowfish/CBC/PKCS5Padding",     P11Cipher,
                 m(CKM_BLOWFISH_CBC));
+
+        d(CIP, "ChaCha20-Poly1305",             P11AEADCipher,
+                s(),
+                m(CKM_CHACHA20_POLY1305));
+
 
         d(CIP, "RSA/ECB/PKCS1Padding",          P11RSACipher,   s("RSA"),
                 m(CKM_RSA_PKCS));
@@ -1165,7 +1181,8 @@ public final class SunPKCS11 extends AuthProvider {
             } else if (type == CIP) {
                 if (algorithm.startsWith("RSA")) {
                     return new P11RSACipher(token, algorithm, mechanism);
-                } else if (algorithm.endsWith("GCM/NoPadding")) {
+                } else if (algorithm.endsWith("GCM/NoPadding") ||
+                           algorithm.startsWith("ChaCha20-Poly1305")) {
                     return new P11AEADCipher(token, algorithm, mechanism);
                 } else {
                     return new P11Cipher(token, algorithm, mechanism);
@@ -1218,6 +1235,8 @@ public final class SunPKCS11 extends AuthProvider {
                     return new sun.security.util.ECParameters();
                 } else if (algorithm == "GCM") {
                     return new sun.security.util.GCMParameters();
+                } else if (algorithm == "ChaCha20-Poly1305") {
+                    return new ChaCha20Poly1305Parameters(); // from SunJCE
                 } else {
                     throw new NoSuchAlgorithmException("Unsupported algorithm: "
                             + algorithm);
--- a/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
+++ b/src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
@@ -632,7 +632,7 @@ public final class SunPKCS11 extends AuthProvider {
 
         d(AGP, "ChaCha20-Poly1305",
                 "com.sun.crypto.provider.ChaCha20Poly1305Parameters",
-                s("1.2.840.113549.1.9.16.3.18", "CHACHA20-POLY1305"),
+                s("1.2.840.113549.1.9.16.3.18", "OID.1.2.840.113549.1.9.16.3.18"),
                 m(CKM_CHACHA20_POLY1305));
 
         d(KA, "DH",             P11KeyAgreement,        s("DiffieHellman"),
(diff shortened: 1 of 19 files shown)


src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -709,6 +720,11 @@ private static void register(Descriptor d) {
         d(CIP, "Blowfish/CBC/PKCS5Padding",     P11Cipher,
                 m(CKM_BLOWFISH_CBC));
 
+        d(CIP, "ChaCha20-Poly1305",             P11AEADCipher,
+                s(),
+                m(CKM_CHACHA20_POLY1305));
+
+
```
TheRealMDoerr, line 727: extra newline

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -626,6 +630,11 @@ private static void register(Descriptor d) {
         d(AGP, "GCM",            "sun.security.util.GCMParameters",
                 m(CKM_AES_GCM));
 
+        d(AGP, "ChaCha20-Poly1305",
+                "com.sun.crypto.provider.ChaCha20Poly1305Parameters",
+                s(),
```
TheRealMDoerr, line 635: aliases should be taken from KnowsOIDs: CHACHA20_POLY1305("1.2.840.113549.1.9.16.3.18", "CHACHA20-POLY1305"),

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -709,6 +720,11 @@ private static void register(Descriptor d) {
         d(CIP, "Blowfish/CBC/PKCS5Padding",     P11Cipher,
                 m(CKM_BLOWFISH_CBC));
 
+        d(CIP, "ChaCha20-Poly1305",             P11AEADCipher,
+                s(),
```
TheRealMDoerr, line 724: like above

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/wrapper/CK_SALSA20_CHACHA20_POLY1305_PARAMS.java
```diff
@@ -0,0 +1,80 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        sb.append(Constants.INDENT);
+        sb.append("Nonce: ");
+        if (nonce == null) {
+            sb.append("null");
+        } else {
+            sb.append("0x");
+            for (byte b: nonce) {
+                sb.append(String.format("0%02X", b));
```
TheRealMDoerr, line 64: other code uses `String.format("%02x", b & 0xff)`. I believe it gets converted to int with sign extend otherwise. Did you check the output?

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/wrapper/CK_SALSA20_CHACHA20_POLY1305_PARAMS.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        sb.append(Constants.NEWLINE);
+        sb.append(Constants.INDENT);
+        sb.append("AAD: ");
+        if (aad == null) {
+            sb.append("null");
+        } else {
+            sb.append("0x");
+            for (byte b: aad) {
+                sb.append(String.format("0%02X", b));
```
TheRealMDoerr, line 75: like above

TheRealMDoerr (changes requested):
Hi Götz, thanks for doing this large backport. Looks basically good, but I have a couple of requests. The small test code refactoring makes sense.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
GoeLin: Fixed.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
GoeLin: Both fixed.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -626,6 +630,11 @@ private static void register(Descriptor d) {
         d(AGP, "GCM",            "sun.security.util.GCMParameters",
                 m(CKM_AES_GCM));
 
+        d(AGP, "ChaCha20-Poly1305",
+                "com.sun.crypto.provider.ChaCha20Poly1305Parameters",
+                s("1.2.840.113549.1.9.16.3.18", "CHACHA20-POLY1305"),
```
TheRealMDoerr, line 635: I've looked a bit more through the code and it seems like the algorithm name is compared case insensitive (`algorithm = algorithm.toUpperCase(Locale.ENGLISH);` in P11SecretKeyFactory.java). So, it seems like "CHACHA20-POLY1305" is redundant because it matches the original name (case insensitive). Other places have an extra entry starting with "OID". I'd use: `s("1.2.840.113549.1.9.16.3.18", "OID.1.2.840.113549.1.9.16.3.18")` which looks consistent with the remaining code. Maybe another reviewer can confirm this.

src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/SunPKCS11.java
```diff
@@ -709,6 +720,10 @@ private static void register(Descriptor d) {
         d(CIP, "Blowfish/CBC/PKCS5Padding",     P11Cipher,
                 m(CKM_BLOWFISH_CBC));
 
+        d(CIP, "ChaCha20-Poly1305",             P11AEADCipher,
+                s("1.2.840.113549.1.9.16.3.18", "CHACHA20-POLY1305"),
+                m(CKM_CHACHA20_POLY1305));
```
TheRealMDoerr, line 725: like above

TheRealMDoerr (approved):
Please consider my proposal. Otherwise, good. Thanks for the backport work!

GoeLin:
Hi, I would appreciate a second review. Does anyone have an opinion on the aliases?
@martinuy If I remember correctly you have made changes to the security files in the past.
Could you have a look?  Or @shipilev or @zhengyu123, would you find time? Thanks!

shipilev:
> Hi, I would appreciate a second review. Does anyone have an opinion on the aliases? @martinuy If I remember correctly you have made changes to the security files in the past. Could you have a look? Or @shipilev or @zhengyu123, would you find time? Thanks!

@martinuy is our security expert, please wait for him to reply :)

TheRealMDoerr:
For the 2nd reviewer: I had proposed to use `s("1.2.840.113549.1.9.16.3.18", "OID.1.2.840.113549.1.9.16.3.18")` to be more consistent with jdk11. Please check.

GoeLin:
@martinuy, will you find time to look at this change this week? Next Tuesday rampdown begins, would be nice to have this pushed by then. 
Thanks, Goetz.

GoeLin:
Hi @shipilev, @TheRealMDoerr 
I would like to push this today before the start of rampdown.
@martinuy did not react yet.  What do you think, is this ok to go anyways?
I think we should not miss it in 11.0.15.  In case the aliases are not good, we can still make 
a follow up change.

TheRealMDoerr:
I'd prefer to push it after updating the aliases. Some jdk11 code may use the "OID." prefix and the algorithm would not be found. I agree with that this backport should go into 11.0.15. Ideally before ramp down or alternatively as critical backport.

theRealAph:
> I'd prefer to push it after updating the aliases. Some jdk11 code may use the "OID." prefix and the algorithm would not be found. I agree with that this backport should go into 11.0.15. Ideally before ramp down or alternatively as critical backport.

I agree too, but let's wait for a comment from Martin.

GoeLin:
> I agree too, but let's wait for a comment from Martin.
You mean Martin Balao, right? He didn't react for 18 days now.
I'd rather have it pushed so that it reaches any downstream testing, 
than wait and do a critical push on last notice.

I updated the aliases as recommended by Martin Doerr.

TheRealMDoerr (approved):
Thanks for the update. Good to go IMHO.

theRealAph:
> > I agree too, but let's wait for a comment from Martin.
> 
> You mean Martin Balao, right? He didn't react for 18 days now. 

I'm sure he never saw it. He's working on it at the moment.

martinuy:
Hi all,

I'm reviewing the backport now and will provide my comments today.

Martin.-

GoeLin:
> I'm sure he never saw it. He's working on it at the moment.

Great, thank you!
Reaching out to somebody is a bit problematic since github: Too many channels (PR, JBS, Mail), and too many mail notifications.  I missed things, too.

martinuy:
Hi @GoeLin,

Thanks for proposing this backports. Looks good to me and I agree with all the previous review comments and fixes.

One minor thing: it looks to me that HexToBytes does not exactly reflect HexFormat::parseHex behavior when it comes to handling null strings. While HexFormat::parseHex requires a non-null value (or throws a NPE) [1], HexToBytes will return an array of length 0. Returning an array of length 0, for HexFormat::parseHex, is reserved for the case in which the string is empty (non-null) only. There might be some performance differences too. Given that this is only used for tests, I believe that no changes are needed to the current PR. Something I was thinking is that HexFormat looks pretty self-contained so I wondered if it is worth moving it to an internal package of java.base (which could be for example jdk.internal.util) and exporting it to modules (such as jdk.crypto.cryptoki) on demand as backports use it. This would allow to use it even beyond SunPKCS11 in the future.

Martin.-

--
[1] - https://github.com/openjdk/jdk/blob/jdk-19+11/src/java.base/share/classes/java/util/HexFormat.java#L527

GoeLin:
Thanks Martin for looking at this so ad-hoc!
/integrate

PR #810, opened 2022-02-08
Backport 53f1b93881a6404508628388ba590fe1f7f958ff

VladimirKempik:
The backport is not applying clean due to the context code difference.
It's needed mostly to resolve this compilation error on macos-aarch64
splashscreen_sys.m:274:39: error: 'NSAlphaFirstBitmapFormat' is deprecated: first deprecated in macOS 10.14 [-Werror,-Wdeprecated-declarations]

### DIFF
--- a/make/autoconf/flags.m4
+++ b/make/autoconf/flags.m4
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2011, 2018, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2011, 2021, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


phohensee (approved):
Lgtm.

Note to Maintainers: 10.9 through 10.14 are eol, and the oldest machines supported by 10.12 are from 2009/2010, see https://support.apple.com/kb/SP742?locale=en_USh. Imo it's safe to say that it's very unlikely there are machines running OSX versions earlier than 10.12.

VladimirKempik:
Thanks Paul.

lewurm:
> Note to Maintainers: 10.9 through 10.14 are eol, and the oldest machines supported by 10.12 are from 2009/2010, see https://support.apple.com/kb/SP742?locale=en_USh. Imo it's safe to say that it's very unlikely there are machines running OSX versions earlier than 10.12.

While I agree, why take the risk?  This can be easily set conditionally for aarch64.

VladimirKempik:
> > Note to Maintainers: 10.9 through 10.14 are eol, and the oldest machines supported by 10.12 are from 2009/2010, see https://support.apple.com/kb/SP742?locale=en_USh. Imo it's safe to say that it's very unlikely there are machines running OSX versions earlier than 10.12.
> 
> While I agree, why take the risk? This can be easily set conditionally for aarch64.

This was reasoned in the bug description:
>Having a large gap between the target versions becomes problematic as we hit a lot of deprecation warnings in shared code. To be able to fix these deprecation warnings, we need a smaller version gap.

lewurm (approved):

VladimirKempik:
@GoeLin, two weeks later - no objections ( https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2022-February/012026.html )
can we get approval or as an alternative, I can reduce this fix and not change min macos version for intel64.
I would like to make macos-aarch64 port buildable without --disable-warnings-as-errors before rdp2.

lewurm:
Can this be merged?

lewurm:
Thank you Paul and Vladimir!

PR #823, opened 2022-02-14
Backport 58dae60da0711c4ae0cb23f8ce2328e051d603b2

alexeybakhtin:
Backport of JDK-8274524 to JDK11u-dev
Backport is almost clean except for a trivial fix for Pattern Matching in the instanceof (not available in JDK11) and braces for the if-clauses.
sun/security/ssl/SSLSocketImpl tests passed


phohensee (approved):
Lgtm.

PR #927, opened 2022-03-21
8228609: G1 copy cost prediction uses used vs. actual copied bytes

linade:
I would like to backport 8228609 which fixes a prediction error regarding cost per bytes copied.

The patch does not apply cleanly, so this backport is basically changing the calculation of copied bytes.

https://bugs.openjdk.java.net/browse/JDK-8227442 is not in 11u. So this patch also uses G1ParScanThreadState::surviving_young_words() in place of _surviving_young_words to adjust for the correct cset index.

---------
### Progress
- [ ] Change must be properly reviewed

### Issue
 * [JDK-8228609](https://bugs.openjdk.java.net/browse/JDK-8228609): G1 copy cost prediction uses used vs. actual copied bytes

### Reviewers
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/927/head:pull/927` \
`$ git checkout pull/927`

`$ git checkout pull/927` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/927/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 927`

`$ git pr show -t 927`

</details>

<a href="https://git.openjdk.java.net/jdk11u-dev/pull/927.diff">https://git.openjdk.java.net/jdk11u-dev/pull/927.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
+++ b/src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
@@ -82,16 +82,19 @@ G1ParScanThreadState::G1ParScanThreadState(G1CollectedHeap* g1h, uint worker_id,
 }
 
 // Pass locally gathered statistics to global state.
-void G1ParScanThreadState::flush(size_t* surviving_young_words) {
+size_t G1ParScanThreadState::flush(size_t* surviving_young_words) {
   _dcq.flush();
   // Update allocation statistics.
   _plab_allocator->flush_and_retire_stats();
   _g1h->g1_policy()->record_age_table(&_age_table);
 
+  size_t sum = 0;
   uint length = _g1h->collection_set()->young_region_length();
   for (uint region_index = 0; region_index < length; region_index++) {
-    surviving_young_words[region_index] += _surviving_young_words[region_index];
+    surviving_young_words[region_index] += G1ParScanThreadState::surviving_young_words()[region_index];
+    sum += G1ParScanThreadState::surviving_young_words()[region_index];
   }
+  return sum;
 }
 
 G1ParScanThreadState::~G1ParScanThreadState() {
@@ -339,6 +342,7 @@ const size_t* G1ParScanThreadStateSet::surviving_young_words() const {
 void G1ParScanThreadStateSet::flush() {
   assert(!_flushed, "thread local state from the per thread states should be flushed once");
 
+  size_t copied_words = 0;
   for (uint worker_index = 0; worker_index < _n_workers; ++worker_index) {
     G1ParScanThreadState* pss = _states[worker_index];
 
@@ -346,10 +350,11 @@ void G1ParScanThreadStateSet::flush() {
       continue;
     }
 
-    pss->flush(_surviving_young_words_total);
+    copied_words += pss->flush(_surviving_young_words_total);
     delete pss;
     _states[worker_index] = NULL;
   }
+  _g1h->g1_policy()->record_copied_bytes(copied_words << LogBytesPerWord);
   _flushed = true;
 }
 
--- a/src/hotspot/share/gc/g1/g1Policy.cpp
+++ b/src/hotspot/share/gc/g1/g1Policy.cpp
@@ -60,6 +60,7 @@ G1Policy::G1Policy(STWGCTimer* gc_timer) :
   _reserve_factor((double) G1ReservePercent / 100.0),
   _reserve_regions(0),
   _rs_lengths_prediction(0),
+  _copied_bytes(0),
   _initial_mark_to_mixed(),
   _collection_set(NULL),
   _g1h(NULL),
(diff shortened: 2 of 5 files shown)


TheRealMDoerr:
This PR is not recognized as backport. Please fix that (e.g. by changing the title to "Backport <hash of original commit>").

linade:
@GoeLin I enabled the tests. Some of them seem to be blocked by a gcc dependency issue. Would you take a look?

GoeLin:
Hi, you should only label a change jdk11u-fix-request once it is ready for integration. I'll remove the label, please add it again if you have a review.

linade:
Gentle ping. Can anyone review this?

The reason I believe it's worth a backport is, apart from what's mentioned in https://bugs.openjdk.java.net/browse/JDK-8228609, it correctly accounts for early reclaimed humongous objects. I think originally the oversight of the size of early reclaimed humongous objects cause the calculation to be erroneous, resulting in a very high copy time prediction and unstable gc frequency.

src/hotspot/share/gc/g1/g1Policy.cpp
```diff
@@ -667,10 +668,9 @@ void G1Policy::record_collection_pause_end(double pause_time_ms, size_t cards_sc
 
     size_t freed_bytes = heap_used_bytes_before_gc - cur_used_bytes;
```
TheRealMDoerr, line 669: `freed_bytes` is now unused.

TheRealMDoerr (commented):
This change is very different from the original one, but fixing it in 11u sounds feasible to me. We'll test it.
Do you have a specific test case which allows observing unstable gc frequency?

linade:
> This change is very different from the original one, but fixing it in 11u sounds feasible to me. We'll test it. Do you have a specific test case which allows observing unstable gc frequency?

We observed this in an application. With additional logging, we can observe the miscalculation from, e.g., test/hotspot/jtreg/gc/g1/TestEagerReclaimHumongousRegions.java

The current way to get copied_bytes will result in negative value. I had
```
..
[0.708s][error][gc] GC(186) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.714s][error][gc] GC(188) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.720s][error][gc] GC(190) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.726s][error][gc] GC(192) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.732s][error][gc] GC(194) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.738s][error][gc] GC(196) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
[0.744s][error][gc] GC(198) cset.used=14641416, freed_bytes=40855816, copied_bytes=-26214400, actual_copied_bytes=0
```
copied_bytes is negative because it is deduced from ``size_t copied_bytes = _collection_set->bytes_used_before() - freed_bytes;``, and ``_collection_set->bytes_used_before()`` didn't consider eager-reclaimed humongous, so it appears to be much smaller than ``freed_bytes``.

There currently is an if condition ``if (_collection_set->bytes_used_before() > freed_bytes)`` to prevent we from getting a negative value. So the current code simply doesn't report the data to the analytics.

Not reporting is fine. At worst we lose some data points. But it can also happen that we get a positive, near-zero ``copied_bytes``. It will make the value ``double cost_per_byte_ms = average_copy_time / (double) copied_bytes;`` very big. This causes young generation to shrink.

src/hotspot/share/gc/g1/g1Policy.cpp
linade: Removed. Thanks.

TheRealMDoerr (approved):
Looks technically good and the fix makes sense to me. Our nightly tests haven't found any issues related to it. It extracts the essential fix from the original change which differs a lot: https://github.com/openjdk/jdk/commit/9a4c25731eef42d18ba0c20ff3eb42f32ae90c00
An additional person with GC experience should review it.
Approvers will have to judge how relevant it is. A comment from the original author would be helpful.

rkennke:
I *hink* it makes sense. It is not exactly a backport, but rather a re-write, and I find it difficult to follow and convince myself that it's doing the correct thing. Therefore it seems somewhat risky.

src/hotspot/share/gc/g1/g1ParScanThreadState.cpp
```diff
@@ -82,16 +82,19 @@ G1ParScanThreadState::G1ParScanThreadState(G1CollectedHeap* g1h, uint worker_id,
...
   _dcq.flush();
   // Update allocation statistics.
   _plab_allocator->flush_and_retire_stats();
   _g1h->g1_policy()->record_age_table(&_age_table);
 
+  size_t sum = 0;
   uint length = _g1h->collection_set()->young_region_length();
   for (uint region_index = 0; region_index < length; region_index++) {
-    surviving_young_words[region_index] += _surviving_young_words[region_index];
+    surviving_young_words[region_index] += G1ParScanThreadState::surviving_young_words()[region_index];
```
rkennke, line 94: Was that code using the wrong index before? IOW, it used the surviving_young_words of the wrong regions?

linade: Yes, that's what I think.

So this pr is achieving two things (sorry if it were not clear):

1. fix the index issue: this causes calculation of the total surviving young words to be incorrect. The correct stats are stored in _surviving_young_words[1, .., n] instead of [0, .., n-1]
3. use a precise "copied bytes" instead of inferring from (used() - freed())

PR #993, opened 2022-04-04
8218965: aix:  support xlclang++ in the compiler detection

backwaterred:
This PR represents my effort to transition to xlclang/++ as the compiler on AIX for jdk11u. These changes are required to support changes to the Harfbuzz library version [[1]](https://bugs.openjdk.java.net/browse/JDK-8282588?focusedCommentId=14480395&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14480395). Unfortunately, it was not a simple, clean backport process. After several backports, merge conflicts, and one (very simple) modification, the result builds.

@TheRealMDoerr and @MBaesken have been involved in updating the Harfbuzz version. I suspect this will be of interest to them.

### Testing

Tier1 tests complete and passing as expected on AIX. Tier2 tests are in progress.

### Notes

The following conflicts were observed while backporting.

[8218965: aix: support xlclang++ in the compiler detection](https://github.com/openjdk/jdk11u-dev/commit/0b0a99557e7ed82c043d94f6a99e67e2e2c8c688)

- hb-atomic-private.hh was modified by the backported commit, but not present in the source. I discarded these changes.

[8224214: [AIX] Remove support for legacy xlc compiler](https://github.com/openjdk/jdk11u-dev/commit/3e10d117386666c139d6da879fee5b35f37722c0)

- ptrQueue.hpp The backport added procedure definitions to the hpp file that were not implemented in the cpp file. I removed these.
- oopStorage.hpp The backport added procedure definitions to the hpp file that were not implemented in the cpp file. I removed these.
- globalDefinitions_xlc.hpp  Add version check for xlc 16.

[8218136: minor hotspot adjustments for xlclang++ from xlc16 on AIX](https://github.com/openjdk/jdk11u-dev/commit/b9061fca09a63877602db4493c4b72ba286ce927)

- src/hotspot/share/runtime/arguments.cpp: Remove `UNSUPPORTED_OPTION_NULL(AllocateOldGenAt)` modified by backport but not present in original source.

[8227171: provide function names in native stack trace on aix with xlc16](https://github.com/openjdk/jdk11u-dev/commit/e56023cc31269cb1681e4d5a05150668bb111e0f)

- flags-cflags.m4: Accepted relevant compiler flag additions from backport.

[8227389: Remove unsupported xlc16 compile options on aix](https://github.com/openjdk/jdk11u-dev/commit/5b766a84e12ba8e0f84775a793c89aceb29a52ec)

[...]

PR #999, opened 2022-04-04
Backport only 8227389

backwaterred:
#993 

8227389: Remove unsupported xlc16 compile options on aix

Reviewed-by: clanger, mdoerr

Backport Conflicts:
- make/autoconf/flags-cflags.m4: Minor change. Accept backported
addition of -qstackprotect.


TheRealMDoerr:
This may break building with older xlc. Maybe we need to distinguish. Please check.

PR #998, opened 2022-04-04
8227171: provide function names in native stack trace on aix with xlc16

backwaterred:
#993

Backport of:

8227171: provide function names in native stack trace on aix with xlc16

Reviewed-by: stuefe

Backport Notes:

- flags-cflags.m4: Minor change. One flag (-qstackprotect) expected by backported commit and not removed by it. Resolved by adding -qstackprotect.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8227171](https://bugs.openjdk.java.net/browse/JDK-8227171): provide function names in native stack trace on aix with xlc16

### Reviewers
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk11u-dev pull/998/head:pull/998` \
`$ git checkout pull/998`

Update a local copy of the PR: \
`$ git checkout pull/998` \
`$ git pull https://git.openjdk.java.net/jdk11u-dev pull/998/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 998`

View PR using the GUI difftool: \
`$ git pr show -t 998`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jdk11u-dev/pull/998.diff">https://git.openjdk.java.net/jdk11u-dev/pull/998.diff</a>

</details>

PR #1030, opened 2022-04-11
Backport 74f9ec23e3eed594177743cbfcbe17705ccce9c1

backwaterred:
Discovered while bringing xlc 16.1 support to jdk11. This change adds stack protection to AIX and cleans up the flags to apply #998.

Reviewed-by: clanger, sgroeger

Backport Conflicts:
- minor conflict in flags-cflags.m4 (potentially caused by comment
addition?)


TheRealMDoerr:
Please try to keep backports as clean as possible. Comments shouldn't be added in backports. That should be done in jdk master.

backwaterred:
The pre-test build failures for x64 Linux should not be caused by this change, and yet I have seen them show up on a few of my backport PRs. I've tried to replicate them and was unable. Should I be concerned, or is it safe to assume they are caused by some other issue?

TheRealMDoerr:
Right, this PR only affects AIX. GHA tests have other problems. Thanks for the update! It's now clean and doesn't need a review.

PR #1056, opened 2022-04-28
8284102: [TESTBUG] Retroactively add regression test for JDK-8272124

jerboaa:
This adds a regression test for the fix originally backported when no cgroups v2 support was present (and, thus, test infra wasn't there either). The test passes. If the fix regets reverted it fails (as expected).


gnu-andrew (approved):
Ok, good to go then.

jerboaa:
Thanks!

PR #1102, opened 2022-05-20
Backport 892d998587a0a7a7ed17586f96ae7dc5b236adb2

TheRealMDoerr:
Backport of [JDK-8139046](https://bugs.openjdk.java.net/browse/JDK-8139046)
Doesn't apply cleanly:

- compile.hpp: Had to take `should_print` from https://github.com/openjdk/jdk/commit/75e9d0a2901c149fc0acf3ed042ddedc689d8d23
- compilerDefinitions.cpp, compile.cpp, idealGraphPrinter.cpp, idealGraphPrinter.hpp: Had to do resolve hunks manually.

### DIFF
--- a/src/hotspot/share/compiler/compilerDefinitions.cpp
+++ b/src/hotspot/share/compiler/compilerDefinitions.cpp
@@ -439,8 +439,8 @@ void CompilerConfig::ergo_initialize() {
   if (!IncrementalInline) {
     AlwaysIncrementalInline = false;
   }
-  if (PrintIdealGraphLevel > 0) {
-    FLAG_SET_ERGO(bool, PrintIdealGraph, true);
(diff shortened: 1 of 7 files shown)


navyxliu (approved):
This code change looks good to me.  I am not a reviewer. we still need other reviewers to approve it. 

This patch is for developer-oriented and only impacts debug-build. Is it necessary to backport to jdk11u? I think all developers of Hotspot/C2 have moved to jdk-tip.

TheRealMDoerr:
Thanks for the review! We want to use it for debugging C2 crashes. Some problems don't reproduce with later JDK releases. That's why we need this fix in 11u.

TheRealMDoerr:
Thanks for the review and the approval!
/integrate

PR #1164, opened 2022-06-21
8288865: [aarch64] LDR instructions must use legitimized addresses

apavlyutkin:
Hi!

Please review the fix to non-volatile memory access on aarch64 that makes LDR instructions to use legitimized addresses exactly like it was done to STR instruction by https://bugs.openjdk.org/browse/JDK-8235385. 

Verification (18.04.6 LTS / aarch64):

test/hotspot/jtreg/compiler/unsafe/NonVolatileMemoryAccessWithLongOffset.java

passes 10 of 10 runs. At least a half of runs failed before the fix.

Regression (18.04.6 LTS / aarch64): hotspot_compiler

@theRealAph @sandlerwang could you look at that? Thank you


apavlyutkin:
ping

PR #1502, opened 2022-11-04
8270947: AArch64: C1: use zero_words to initialize all objects

navyxliu:
We would like to backport [JDK-8270947](https://bugs.openjdk.org/browse/JDK-8270947) to jdk11u. It's not only less code to maintain, but also better generated code. C1 can leverage DC ZVA in large allocations on aarch64. The secondary consideration is that openjdk11u aligns with oraclejdk11u. 

This isn't a clean backport. Small cosmetic changes are made in `MacroAssembler::zero_words(Register ptr, Register cnt)` chunk. 

There are also 3 trailing bugfixes. We will post them together as stacked PRs

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8270947](https://bugs.openjdk.org/browse/JDK-8270947): AArch64: C1: use zero_words to initialize all objects

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev pull/1502/head:pull/1502` \
`$ git checkout pull/1502`

Update a local copy of the PR: \
`$ git checkout pull/1502` \
`$ git pull https://git.openjdk.org/jdk11u-dev pull/1502/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1502`

View PR using the GUI difftool: \
`$ git pr show -t 1502`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk11u-dev/pull/1502.diff">https://git.openjdk.org/jdk11u-dev/pull/1502.diff</a>

</details>

PR #1503, opened 2022-11-04
Backport 0948c097a855dcc9a8a437b4618d7c1922722eab

navyxliu:
This is a stacked PR(2/3) of https://github.com/openjdk/jdk11u-dev/pull/1502. 

This patch is one of the trailing bugfix of JDK-8270947. It's not a clean patch. I just substitute 'nullptr' with NULL because jdk11u is using gnu++11 to compile HotSpot.


phohensee (approved):
Lgtm.

PR #1510, opened 2022-11-09
8296652: Restore windows aarch64 fixpath patch that was removed in 8239708

gdams:
When https://bugs.openjdk.org/browse/JDK-8239708 was backported to 11u, the Windows aarch64 specific fixpath patch added in https://github.com/openjdk/jdk11u/commit/7bdb8ac7f9f9a83779427da7a6d7f5e9ae548db7#diff-cbf3a26a7e5854d626b3c7e11c8bcb3c287ed95113360aaf220108bfa018e066was missed out. This will need restoring


RealCLanger (approved):
This looks good. Thanks for restoring.

RealCLanger:
@erikj79, any idea why @gdams is not recognized as committer here? As per the census he should be. Anyway...

/sponsor

PR #1639, opened 2023-01-04
8299616: [11u] Bootcycle build fails after JDK-8257679 backport

shipilev:
Currently, JDK 11u `make bootcycle-images` fails with:

```
 bootcycle-spec.gmk:32: *** non-numeric first argument to `word' function: ''. Stop.
```

The apparent reason is [JDK-8257679](https://bugs.openjdk.org/browse/JDK-8257679) that did the following change:
 https://github.com/openjdk/jdk11u-dev/commit/40f4fc2da8532252a1660c7f082eb19046f4bc70#diff-935527b393cb4625a8874febf5fe39d43479a2228ef437f4e104ba3001e1e30fL31-L40

Note that the left-over hunk uses `JAVA_EXEC_POS`, which is defined in the hunk that was removed. This PR reinstantiates that hunk.

Additional testing:
 - [x] Linux x86_64 fastdebug `make bootcycle-images`
 - [x] Windows x86_64 fastdebug `make bootcycle-images`

---------
### Progress

### Issue
 * [JDK-8299616](https://bugs.openjdk.org/browse/JDK-8299616): [11u] Bootcycle build fails after JDK-8257679 backport

### Reviewers
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**) ⚠️ Review applies to [b0ad92ed](https://git.openjdk.org/jdk11u-dev/pull/1639/files/b0ad92eda57ad20649d46d0654e48f5e534f9202)
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**) ⚠️ Review applies to [b0ad92ed](https://git.openjdk.org/jdk11u-dev/pull/1639/files/b0ad92eda57ad20649d46d0654e48f5e534f9202)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev pull/1639/head:pull/1639` \
`$ git checkout pull/1639`

`$ git checkout pull/1639` \
`$ git pull https://git.openjdk.org/jdk11u-dev pull/1639/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1639`

`$ git pr show -t 1639`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/1639.diff">https://git.openjdk.org/jdk11u-dev/pull/1639.diff</a>

</details>

### DIFF
--- a/make/autoconf/bootcycle-spec.gmk.in
+++ b/make/autoconf/bootcycle-spec.gmk.in
@@ -28,6 +28,14 @@
 # First include the real base spec.gmk file
 include @SPEC@
 
+# Check that the user did not try to specify a different java to use for compiling.
+# On windows we need to account for fixpath being first word.
+ifeq ($(firstword $(JAVA)),$(FIXPATH))
+  JAVA_EXEC_POS=2
+else
+  JAVA_EXEC_POS=1
+endif
+
 # Override specific values to do a boot cycle build
 ifneq ($(word $(JAVA_EXEC_POS),$(SJAVAC_SERVER_JAVA)),$(word $(JAVA_EXEC_POS),$(JAVA)))
   $(error Bootcycle builds are not possible if --with-sjavac-server-java is specified)
--- a/make/autoconf/bootcycle-spec.gmk.in
+++ b/make/autoconf/bootcycle-spec.gmk.in
@@ -35,11 +35,11 @@ ifeq ($(firstword $(JAVA)),$(FIXPATH))
 else
   JAVA_EXEC_POS=1
 endif
-
-# Override specific values to do a boot cycle build
 ifneq ($(word $(JAVA_EXEC_POS),$(SJAVAC_SERVER_JAVA)),$(word $(JAVA_EXEC_POS),$(JAVA)))
   $(error Bootcycle builds are not possible if --with-sjavac-server-java is specified)
 endif
+
+# Override specific values to do a boot cycle build
 
 # Use a different Boot JDK
 BOOT_JDK := $(JDK_IMAGE_DIR)


gnu-andrew (approved):
Thanks for tracking this down so quickly. After you identified the problematic patches, I came to the same conclusion that this is down to `JAVA_EXEC_POS` still being used in `bootcycle-spec.gmk.in`, but now undefined.

I think the reason for the change was two-fold:

1. `$(FIXPATH)` was removed from `JAVA` in `spec.gmk.in` and only added to `JAVA_CMD` in `bootcycle-spec.gmk.in` so the second branch would always be true.
2. The block as a whole was removed as `JAVA_EXEC_POS` is unused after JDK-8244036 (not backported to 11u)

The safest option is to restore what we had before, as your patch does, but it may be that `JAVA_EXEC_POS` can now always be 1 (or just removed and a hardcoded 1 used in `bootcycle-spec.gmk.in`.

I'm happy with this as the lowest risk fix, assuming Windows builds fine.

RealCLanger:
Ah, yes, [JDK-8244036](https://bugs.openjdk.org/browse/JDK-8244036), is not in jdk11u, so removing this block was wrong. Thanks for finding & fixing. Please integrate this and I'll cherry-pick it into the bundle for 11.0.18, too.

RealCLanger (approved):

shipilev:
> The safest option is to restore what we had before, as your patch does, but it may be that `JAVA_EXEC_POS` can now always be 1 (or just removed and a hardcoded 1 used in `bootcycle-spec.gmk.in`).

True. But, I could not convince myself there is a guarantee for `fixpath` to be absent in either `$SJAVAC_SERVER_JAVA` or `$JAVA`, so kept the most pessimistic fix.

I did massage the comments a bit, so that they would match the original code shape.

shipilev:
Windows and Linux bootcycle builds are fine. GHAs are clean. Push approval is there. I am integrating.

/integrate

gnu-andrew:
> > The safest option is to restore what we had before, as your patch does, but it may be that `JAVA_EXEC_POS` can now always be 1 (or just removed and a hardcoded 1 used in `bootcycle-spec.gmk.in`).
> 
> True. But, I could not convince myself there is a guarantee for `fixpath` to be absent in either `$SJAVAC_SERVER_JAVA` or `$JAVA`, so kept the most pessimistic fix.
> 
> I did massage the comments a bit, so that they would match the original code shape.

Yes, that was my thinking too. Better safe than sorry.

PR #1799, opened 2023-03-10
Backport f07dcf471c160e09fbc748124998923e7d453e66

Autumn808:
This backport is part of the 28 backport Accessibility series JDK-8152350. Commented out line 61 of commoneComponentAccesssibilty.m until backport JDK-8262981 is reworked.

### DIFF
--- a/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/CommonComponentAccessibility.m
+++ b/src/java.desktop/macosx/native/libawt_lwawt/awt/a11y/CommonComponentAccessibility.m
@@ -48,7 +48,7 @@ static jobject sAccessibilityClass = NULL;
     /*
      * Here we should keep all the mapping between the accessibility roles and implementing classes
      */
-    rolesMap = [[NSMutableDictionary alloc] initWithCapacity:26];
+    rolesMap = [[NSMutableDictionary alloc] initWithCapacity:34];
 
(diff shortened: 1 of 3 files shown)


phohensee (approved):
Clean except for the context difference in CommonComponentAccessibility.m. Lgtm.

GoeLin:
Hi @autumn808 @phohensee
you say you had to work around 8262981. What stops that change from being pushed? Can't you move the CSR to proposed?

Autumn808:
> Hi @Autumn808 @phohensee you say you had to work around 8262981. What stops that change from being pushed? Can't you move the CSR to proposed?

we didn’t see a CSR for the Oracle backport, so we believe they adopted a different approach which we’re figuring out how to duplicate. The CSR looks like it changes a public interface, which would in theory require a JDK 11 update JSR.

RealCLanger:
> > Hi @Autumn808 @phohensee you say you had to work around 8262981. What stops that change from being pushed? Can't you move the CSR to proposed?
> 
> @GoeLin we didn’t see a CSR for the Oracle backport, so we believe they adopted a different approach which we’re figuring out how to duplicate. The CSR looks like it changes a public interface, which would in theory require a JDK 11 update JSR.

Right, interface changes can't be backported. Looking at the change 8262981, you could probably just backport the native code to keep the codebase even. Would you mind doing that?

PR #2035, opened 2023-07-10
8229147: Linux os::create_thread() overcounts guardpage size with newer glibc (>=2.27)

RealLucy:
This backport is for parity with 11.0.21-oracle. 
Fix didn't apply at all, os/linux/os_linux.cpp was too different in particular. 
I managed to create a working version by massaging and applying the changes manually. 

There is some risk involved. It is not caused by the intentional change of behavior but by potential collateral damage. 

Tests (SAP nightly test suite) do not indicate any issues.
All the additional commits were as well tested overnight with no issues found.
Reviews received.

---------
### Progress

### Issue
 * [JDK-8229147](https://bugs.openjdk.org/browse/JDK-8229147): Linux os::create_thread() overcounts guardpage size with newer glibc (&gt;=2.27) (**Bug** - P4)

### Reviewers
 * [Goetz Lindenmaier](https://openjdk.org/census#goetz) (@GoeLin - **Reviewer**)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/2035/head:pull/2035` \
`$ git checkout pull/2035`

`$ git checkout pull/2035` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/2035/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2035`

`$ git pr show -t 2035`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/2035.diff">https://git.openjdk.org/jdk11u-dev/pull/2035.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/2035#issuecomment-1629636665)

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -908,8 +908,9 @@ bool os::Linux::adjustStackSizeForGuardPages() {
   return _adjustStackSizeForGuardPages;
 }
 
-#if defined(__GLIBC__) // TLS not in jdk11
-static bool AdjustStackSizeForTLS = false; // Dumy decl as substitute for cmdline parameter
+#if defined(__GLIBC__)
+// Dummy decl as substitute for cmdline parameter. TLS not in jdk11.
+static const bool AdjustStackSizeForTLS = false;
 static void init_adjust_stacksize_for_guard_pages() {
   assert(_get_minstack_func == NULL, "initialization error");
   _get_minstack_func =(GetMinStack)dlsym(RTLD_DEFAULT, "__pthread_get_minstack");
@@ -966,10 +967,13 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   // Apply stack size adjustments if needed. However, be careful not to end up
   // with a size of zero due to overflow. Don't add the adjustment in that case.
   size_t stack_adjust_size = 0;
+#if defined(__GLIBC__)
   if (AdjustStackSizeForTLS) {
     // Adjust the stack_size for on-stack TLS - see get_static_tls_area_size().
     stack_adjust_size += get_static_tls_area_size(&attr);
-  } else if (os::Linux::adjustStackSizeForGuardPages()) {
+  } else
+#endif
+  if (os::Linux::adjustStackSizeForGuardPages()) {
     stack_adjust_size += guard_size;
   }
 
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -908,9 +908,10 @@ bool os::Linux::adjustStackSizeForGuardPages() {
   return _adjustStackSizeForGuardPages;
 }
 
-#if defined(__GLIBC__)
 // Dummy decl as substitute for cmdline parameter. TLS not in jdk11.
 static const bool AdjustStackSizeForTLS = false;
+
+#if defined(__GLIBC__)
 static void init_adjust_stacksize_for_guard_pages() {
   assert(_get_minstack_func == NULL, "initialization error");
   _get_minstack_func =(GetMinStack)dlsym(RTLD_DEFAULT, "__pthread_get_minstack");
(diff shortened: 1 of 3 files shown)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -835,6 +835,104 @@ static void *thread_native_entry(Thread *thread) {
...
+// stack_size by the size of the guard pages to mimic proper behaviour.
+// The fix in glibc 2.27 has now been backported to numerous earlier
+// glibc versions so we need to do a dynamic runtime check.
+static bool _adjustStackSizeForGuardPages = true;
+bool os::Linux::adjustStackSizeForGuardPages() {
+  return _adjustStackSizeForGuardPages;
+}
+
+#if defined(__GLIBC__) // TLS not in jdk11
+static bool AdjustStackSizeForTLS = false; // Dumy decl as substitute for cmdline parameter
```
GoeLin, line 912: AdjustStackSizeForTLS is declared under ifdef GLIBC, but
the usage is not. Is this correct?
I think the comment " // TLS not in jdk11" should go into a new line.
Also, maybe we should make this "const" so it can be optimized?

GoeLin (commented):
It is a bummer "8225035: Thread stack size issue caused by large TLS size"
is not in 11. But that was a quite unstable change and has a CSR,
so no good candidate to backport.

This backport contains parts of 8225035, though, but they are blocked
by setting the corresponding flag statically to false.
I think this makes sense as this keeps the code similar
and thus easier to maintai

tstuefe:
I am unsure about this change. I think in itself it's not that important. Effectively it increases stack size of Non-Java-Threads by one page (Java threads should not use glibc guard pages).

Thing is, JDK-8225035 in itself is a lot more important than this change since it affects stack size for running hotspot embedded in a native app that uses a lot of TLS. And that delta is unpredictable, whereas here we are talking about one page. 

But as @GoeLin writes, it is also more risky. And 11 is long out in the field; all native apps did have time getting used to the current VM behavior.

With downporting JDK-8229147, I don't see why we need the TLS parts of JDK-8225035 at all if all we want to have is "+ 1 page if glibc >= 2.27". You even hardcode AdjustStackSizeForTLS to false. Why not leave it out completely?

GoeLin:
This one is a fix that takes effect for everybody. 
8225025 needs to be enabled by setting a flag. I don't think these should be compared.
I assume there is a good reason this made it's way to jdk8 of Oracle and I don't see a reason not to go along. 
The implementation is not that complex that the risk is too high to integrate it.

tstuefe:
> The implementation is not that complex that the risk is too high to integrate it.

Okay. Yes, you are right, its simple.

My only suggestion would be to remove all that TLS stuff, since it is not needed, and if we later want to downport 8225025, it would be more straightforward.

See https://gist.github.com/tstuefe/8a0fd30618f1d0e085b5ca12d7c156cd . Tested, works.

RealLucy:
The new commit reflects the changes requested by @GoeLin 
Without the TLS stuff, this no longer feels like a backport to me. It is a separate fix then, triggered by JDK-8229147. 

I'm only the backport person. You (@tstuefe and @GoeLin) decide and let me know.

GoeLin:
it probably would be more clean if you movee AdjustStackSizeForTLS out from the ifdef glibc, then you don't need the ifdef you just added below.
But as this change does not touch get_static_tls_area_size() nor the block that calls it, I am also fine with removing both. We won't loose parts of this change as I understand.
After all it makes no big difference as the code is dead.

tstuefe:
>But as this change does not touch get_static_tls_area_size() nor the block that calls it, I am also fine with removing both. We won't loose parts of this change as I understand.

I leave it up to Lutz. I think less is better, but he is the patch author so his opinion has preference.

GoeLin (approved):
LGTM. Thanks for the edits!

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -835,6 +835,106 @@ static void *thread_native_entry(Thread *thread) {
...
+// stack_size by the size of the guard pages to mimic proper behaviour.
+// The fix in glibc 2.27 has now been backported to numerous earlier
+// glibc versions so we need to do a dynamic runtime check.
+static bool _adjustStackSizeForGuardPages = true;
+bool os::Linux::adjustStackSizeForGuardPages() {
+  return _adjustStackSizeForGuardPages;
+}
+
+// Dummy decl as substitute for cmdline parameter. TLS not in jdk11.
+static const bool AdjustStackSizeForTLS = false;
```
tstuefe, line 912: Nit: could even be constexpr.

tstuefe (approved):
Looks good!

src/hotspot/os/linux/os_linux.cpp
RealLucy: This is just personal preference: I'd like to keep the integrated fix close to the original. I understand the KISS approach, though. It's just that I'm biased a bit more to the other side this time. 

Very important: thank you for the discussion and for the reviews!

PR #2087, opened 2023-08-17
8284772: GHA: Use GCC Major Version Dependencies Only

gnu-andrew:
GHA is currently broken in 11u-dev as the referenced Ubuntu gcc package, `10.3.0-1ubuntu1~20.04`, is no longer available.

Rather than bumping this yet again, this PR suggests dropping the specific version as we did some time ago in 8u. The requirement still specifies a specific major version of GCC. It just means the dependency isn't broken every time Ubuntu bumps to a new minor release or even just makes a minor change to the package alone.

Note that the current setup does not guarantee sticking with an exact version of GCC anyway, because - as seen by the current GHA breakage - older versions get removed from the package repository. All we get from this exact version requirement is sporadic breakage. If we truly want a static version of GCC, we need to provide our own as we do with the JDK.

---------
### Progress

### Issue
 * [JDK-8284772](https://bugs.openjdk.org/browse/JDK-8284772): GHA: Use GCC Major Version Dependencies Only (**Bug** - P4)

### Reviewers
 * [Sergey Bylokhov](https://openjdk.org/census#serb) (@mrserb - **Reviewer**)
 * [Volker Simonis](https://openjdk.org/census#simonis) (@simonis - **Reviewer**) ⚠️ Review applies to [5ab54a0f](https://git.openjdk.org/jdk11u-dev/pull/2087/files/5ab54a0f9f8c79c704e788a60c6871163d8923c7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/2087/head:pull/2087` \
`$ git checkout pull/2087`

`$ git checkout pull/2087` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/2087/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2087`

`$ git pr show -t 2087`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/2087.diff">https://git.openjdk.org/jdk11u-dev/pull/2087.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/2087#issuecomment-1682871785)

### DIFF
--- a/.github/workflows/build-cross-compile.yml
+++ b/.github/workflows/build-cross-compile.yml
@@ -32,14 +32,6 @@ on:
         required: false
         type: string
         default: '10'
-      apt-gcc-version:
-        required: false
-        type: string
-        default: '10.3.0-1ubuntu1~20.04'
-      apt-gcc-cross-suffix:
-        required: false
-        type: string
-        default: 'cross1'
 
 jobs:
   build-cross-compile:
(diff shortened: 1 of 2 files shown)


mrserb:
maybe we can do the same for 17?

mrserb (approved):

mrserb:
[build-cross-compile.yml](https://github.com/openjdk/jdk11u-dev/blob/master/.github/workflows/build-cross-compile.yml#L38) should be updated as well?

simonis (approved):
Thanks for the quick fix.

Looks good to me.

mrserb (changes requested):
Yeah cross compilation should be updated.

gnu-andrew:
> Yeah cross compilation should be updated.

Good catch. My `grep` didn't catch that as the version is set indirectly as a default value.

For the current 11u version of `build-cross-compile.yml`, we just drop `apt-gcc-version` and `apt-gcc-cross-suffix` altogether as they only relate to the specific version e.g. we now ask for `gcc-10-s390x-linux-gnu` not `gcc-10-s390x-linux-gnu=10.3.0-1ubuntu1~20.04cross1` (https://packages.ubuntu.com/jammy/gcc-10-s390x-linux-gnu)

gnu-andrew:
> maybe we can do the same for 17?

I'm happy to propose it if this goes through. It's not as critical as https://github.com/openjdk/jdk17u/commit/260f28761a87553fd634615df5ba3db5b76ae692 was already added there. The need for things like [JDK-8313428](https://bugs.openjdk.org/browse/JDK-8313428) included in that change is exactly why I think this is the better long-term option.

mrserb (approved):

gnu-andrew:
Thanks for the reviews. I've flagged the bug for 11u approval.

gnu-andrew:
I see `jdk11u-fix-yes`.
/integrate

shipilev:
I don't think we should be doing these point fixes that diverge 11u GHA, and effectively do forward-ports from 8u. If we want to change how minor GCC versions are handled, we need to do this in mainline, and then back-port that change properly to 21u, 17u, 11u. 8u is harder, because GHA is way backwards here, missing major GHA rewrites.

But for 11u we should just keep maintaining the parity with mainline. In other words, do 11u GHA updates like we did for 17u: https://github.com/openjdk/jdk11u-dev/pull/2090. That PR effectively reverts this one.

PR #2090, opened 2023-08-21
8293107: GHA: Bump to Ubuntu 22.04

shipilev:
This similar to the bulk GHA update we did here: https://github.com/shipilev/jdk17u-dev/commit/260f28761a87553fd634615df5ba3db5b76ae692 -- but now for 11u.

Note this reverts and then reapplies [JDK-8284772](https://bugs.openjdk.org/browse/JDK-8284772) to avoid backport conflicts.

There are other GHA improvements we would backport later. This one atomic PR brings in 11u GHA closer to mainline GHA state without breaking the builds. For reference, after this PR, the 11u-vs-mainline GHA diff would be this: [gha-11u-vs-mainline.patch](https://github.com/openjdk/jdk11u-dev/files/12395257/gha-11u-vs-mainline.patch)

---------
### Progress

### Issues
 * [JDK-8293107](https://bugs.openjdk.org/browse/JDK-8293107): GHA: Bump to Ubuntu 22.04 (**Enhancement** - P4)
 * [JDK-8293098](https://bugs.openjdk.org/browse/JDK-8293098): GHA: Harmonize GCC version handling for host and cross builds (**Enhancement** - P4)
 * [JDK-8293361](https://bugs.openjdk.org/browse/JDK-8293361): GHA: dump config.log in case of configure failure (**Bug** - P4)
 * [JDK-8295213](https://bugs.openjdk.org/browse/JDK-8295213): Run GHA manually with user-specified make and configure arguments (**Enhancement** - P4)
 * [JDK-8295885](https://bugs.openjdk.org/browse/JDK-8295885): GHA: Bump gcc versions (**Enhancement** - P4)
 * [JDK-8313428](https://bugs.openjdk.org/browse/JDK-8313428): GHA: Bump GCC versions for July 2023 updates (**Enhancement** - P4)

### Reviewers
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/2090/head:pull/2090` \
`$ git checkout pull/2090`

`$ git checkout pull/2090` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/2090/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2090`

`$ git pr show -t 2090`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/2090.diff">https://git.openjdk.org/jdk11u-dev/pull/2090.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/2090#issuecomment-1686288723)

### DIFF
--- a/.github/workflows/build-cross-compile.yml
+++ b/.github/workflows/build-cross-compile.yml
@@ -29,17 +29,14 @@ on:
   workflow_call:
     inputs:
       gcc-major-version:
-        required: false
+        required: true
         type: string
-        default: '10'
       apt-gcc-version:
-        required: false
+        required: true
         type: string
-        default: '10.3.0-1ubuntu1~20.04'
-      apt-gcc-cross-suffix:
-        required: false
+      apt-gcc-cross-version:
+        required: true
         type: string
-        default: 'cross1'
 
 jobs:
   build-cross-compile:
@@ -95,8 +92,8 @@ jobs:
           sudo apt-get install \
               gcc-${{ inputs.gcc-major-version }}=${{ inputs.apt-gcc-version }} \
               g++-${{ inputs.gcc-major-version }}=${{ inputs.apt-gcc-version }} \
-              gcc-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-version }}${{ inputs.apt-gcc-cross-suffix }} \
-              g++-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-version }}${{ inputs.apt-gcc-cross-suffix }} \
+              gcc-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-cross-version }} \
+              g++-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-cross-version }} \
               libxrandr-dev libxtst-dev libcups2-dev libasound2-dev
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${{ inputs.gcc-major-version }} 100 --slave /usr/bin/g++ g++ /usr/bin/g++-${{ inputs.gcc-major-version }}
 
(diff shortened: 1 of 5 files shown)


tstuefe:
What is the reason for this PR, reducing maintenance by reducing delta to upstream?

I would slightly prefer testing 11u on older systems with older glibcs (20.04 still rocks 2.31, which predates mallinfo2() for instance).

shipilev:
We had the [longer discussion about this already](https://bugs.openjdk.org/browse/JDK-8293107) in the RFE itself. 

> What is the reason for this PR, reducing maintenance by reducing delta to upstream?

Yes.

This is how you look at it: GHA is not under our control. When GHA infra changes, it changes globally for all JDK releases currently in testing. Therefore, the sane way to maintain GHA across JDK release repos is to keep it in sync as much as possible. That's what we did in 17u, and this is what we have to do in 11u, lest somebody is signing up to maintain the diverged 11u GHA going forward.

> I would slightly prefer testing 11u on older systems with older glibcs (20.04 still rocks 2.31, which predates mallinfo2() for instance).

We cannot do toolchain compatibility testing with GHA, because GHA infra is not under our control. If we keep old GHA code, the testing would first slow down, then it would start sporadically break on the way to EOL, and then it would completely break.

mrserb:
Will this change causes the same breakages we had when the minor version of gcc was changed? I prefer to ignore that minor number in GA.

shipilev:
> Will this change causes the same breakages we had when the minor version of gcc was changed? I prefer to ignore that minor number in GA.

In current form, yes, that's a risk that is shared with mainline, 21u, 17u, and until recently 11u GHAs. It would still be there until we backport the GHA change that drops the minor GCC version from all GHAs.

shipilev:
> > Will this change causes the same breakages we had when the minor version of gcc was changed? I prefer to ignore that minor number in GA.
> 
> In current form, yes, that's a risk that is shared with mainline, 21u, 17u, and until recently 11u GHAs. It would still be there until we backport the GHA change that drops the minor GCC version from all GHAs.

Actually, we should just reapply [JDK-8284772](https://bugs.openjdk.org/browse/JDK-8284772) on top, so minor versions get eliminated again. I just approved [JDK-8284772](https://bugs.openjdk.org/browse/JDK-8284772) for inclusion in mainline, which should eliminate the 11u vs mainline difference again.

shipilev:
I think we should proceed with this. Can I have some formal reviews?

RealCLanger (approved):
Looks good, thanks for the effort. Please add approval labels.

shipilev:
> Looks good, thanks for the effort. Please add approval labels.

Thanks! I think I marked all mentioned issues now.

shipilev:
Got the push approval for all issues now. Integrating to make GHA more reliable.

/integrate

PR #2146, opened 2023-09-25
Bacport 6c61bc195090abf73683b811e214810a1226d299

shipilev:
Semi-clean backport to improve JVM diagnostics. The context in 11u is a bit different, but otherwise the patch is the same.


GoeLin:
Pre-submit failure: timeouts in jshell tests. Pretty sure unrelated.

shipilev:
Yes, I think those failures are unrelated. Thanks!

/integrate

PR #2202, opened 2023-10-20
Backport bc2af47e1e4e6e96020e03a60ce098ddd17f63ba

GoeLin:
I backport this for parity with 11.0.20-oracle.

The change in Provider.java had to be reworked.
The basic idea is to not return but keep the result
until it can be passed to the event.

In two other places I did trivial resolves due to context.


MBaesken (approved):
LGTM

PR #2254, opened 2023-11-06
Backport 8026393

GoeLin:
As I understand, this test needs to be removed.


RealLucy (approved):
LGTM.
It is my understanding as well that this test just has to be removed.

PR #2452, opened 2024-01-10
Backport d8af7a6014055295355a1242db6c2872299c6398

parttimenerd:


phohensee (commented):
The 17u backport didn't remove the call to set_in_asgct() in forte.cpp that's removed here. Perhaps 17u should be fixed?

In none of tip, 21u, or 17u is _in_asgct set false in thread.cpp. Is it really needed in 11u?

jbachorik:
@phohensee 
In tip, the `_in_asgct` is initialized in-place and non-static data member initializers are not allowed in JDK 11 codebase.
Considering the style of initialization of data members in thread.hpp I would argue that this is necessary and correct.

jbachorik (approved):

parttimenerd:
Thanks for the comment, I'm on it.

parttimenerd:
/approval request Fixes SIGBUS caused by AsyncGetCallTrace on M1
Clean besides a compilation issue. Tested locally.

phohensee (approved):

PR #2451, opened 2024-01-10
8247818: GCC 10 warning stringop-overflow with symbol code

jmtd:
11u-dev is tricky to build with GCC 11 and newer. GCC 11 is the default version in current Debian stable (and likely derivatives e.g. Ubuntu, but I have not checked) as well as RHEL 9 (and derivatives, likewise unchecked), so I think it's worthwhile to improve that.

This particular patch addresses one of the obstacles (a warning from GCC 10 onwards). It doesn't apply clean, but cherry-pick conflicts were trivial to resolve. It was backported from 16 to 15 and 13 and seems worthwhile for 11u.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] [JDK-8247818](https://bugs.openjdk.org/browse/JDK-8247818) needs maintainer approval
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8247818](https://bugs.openjdk.org/browse/JDK-8247818): GCC 10 warning stringop-overflow with symbol code (**Bug** - P4 - Approved)

### Reviewers
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/2451/head:pull/2451` \
`$ git checkout pull/2451`

Update a local copy of the PR: \
`$ git checkout pull/2451` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/2451/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 2451`

View PR using the GUI difftool: \
`$ git pr show -t 2451`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk11u-dev/pull/2451.diff">https://git.openjdk.org/jdk11u-dev/pull/2451.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/2451#issuecomment-1884840855)

PR #2470, opened 2024-01-22
8324307: [11u] hotspot fails to build with GCC 12 and newer (non-static data member initializers)

jmtd:
GCC 12 and newer warn about non-static data member initialisers when operating under -std=gnu++98 (jdk11u-dev's C++ standards version). These warnings are promoted to errors by default.

This is an issue for jdk11u-dev which uses the C++ standard version '-std=gnu++98'. It is not an issue for later JDK versions which have moved to newer standards (JEP 347 moved JDK16 to -std=c++14): therefore an 11u-specific fix is required.

The approach used in this PR is to remove the initialisers, in common with what we did when backporting the same code to jdk8u-dev.

---------
### Progress
- [x] [JDK-8324307](https://bugs.openjdk.org/browse/JDK-8324307) needs maintainer approval

### Issue
 * [JDK-8324307](https://bugs.openjdk.org/browse/JDK-8324307): [11u] hotspot fails to build with GCC 12 and newer (non-static data member initializers) (**Bug** - P4 - Approved)

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/2470/head:pull/2470` \
`$ git checkout pull/2470`

`$ git checkout pull/2470` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/2470/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2470`

`$ git pr show -t 2470`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/2470.diff">https://git.openjdk.org/jdk11u-dev/pull/2470.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/2470#issuecomment-1903901399)

### DIFF
--- a/src/hotspot/os/linux/cgroupV1Subsystem_linux.hpp
+++ b/src/hotspot/os/linux/cgroupV1Subsystem_linux.hpp
@@ -105,11 +105,11 @@ class CgroupV1Subsystem: public CgroupSubsystem {
 
   private:
     /* controllers */
-    CachingCgroupController* _memory = NULL;
-    CgroupV1Controller* _cpuset = NULL;
-    CachingCgroupController* _cpu = NULL;
-    CgroupV1Controller* _cpuacct = NULL;
-    CgroupV1Controller* _pids = NULL;
+    CachingCgroupController* _memory;
+    CgroupV1Controller* _cpuset;
+    CachingCgroupController* _cpu;
+    CgroupV1Controller* _cpuacct;
+    CgroupV1Controller* _pids;
 
     char * pids_max_val();
 
--- a/src/hotspot/os/linux/cgroupV2Subsystem_linux.hpp
+++ b/src/hotspot/os/linux/cgroupV2Subsystem_linux.hpp
@@ -51,10 +51,10 @@ class CgroupV2Controller: public CgroupController {
 class CgroupV2Subsystem: public CgroupSubsystem {
   private:
     /* One unified controller */
-    CgroupController* _unified = NULL;
+    CgroupController* _unified;
     /* Caching wrappers for cpu/memory metrics */
-    CachingCgroupController* _memory = NULL;
-    CachingCgroupController* _cpu = NULL;
+    CachingCgroupController* _memory;
+    CachingCgroupController* _cpu;
 
     char *mem_limit_val();
     char *mem_swp_limit_val();


jerboaa (approved):
OK.

jmtd:
/approval request to fix building 11u-dev with GCC 12+. 11u-specific fix needed due to later JDKs moving to a newer C++ standards version. TIA!

GoeLin:
@jmtd, if these initializers are not needed at all, shouldn't they be removed in the higher releases as well?

GoeLin:
Removing the tag in the meantime

jmtd:
Hi @GoeLin ,
> @jmtd, if these initializers are not needed at all, shouldn't they be removed in the higher releases as well?

This is a really good question and I took some time to ponder my reply.

They are not needed from a computational POV because the relevant fields are all initialised in the object constructors in all cases. However, to know this I had to read through the class definition to be sure.  I didn't write the original code but if I had I would have likely also preferred an explicit null initialisation in the member declaration, which makes clear the intent that we do not have uninitialised fields and do not risk reading from uninitialised fields. There's a chance the constructor could be altered later and an iniitialisation accidentally removed, or an alternative constructor added lacking the initialisation, etc.

I doubt that anyone would welcome a patch to remove the redundant initialisers in main, but I could propose one if you disagree. WDYT?

gnu-andrew:
> Hi @GoeLin ,
> 
> > @jmtd, if these initializers are not needed at all, shouldn't they be removed in the higher releases as well?
> 
> This is a really good question and I took some time to ponder my reply.
> 
> They are not needed from a computational POV because the relevant fields are all initialised in the object constructors in all cases. However, to know this I had to read through the class definition to be sure. I didn't write the original code but if I had I would have likely also preferred an explicit null initialisation in the member declaration, which makes clear the intent that we do not have uninitialised fields and do not risk reading from uninitialised fields. There's a chance the constructor could be altered later and an iniitialisation accidentally removed, or an alternative constructor added lacking the initialisation, etc.
> 
> I doubt that anyone would welcome a patch to remove the redundant initialisers in main, but I could propose one if you disagree. WDYT?

I would say this is the sort of change that belongs only in 11u and older, for compatibility reasons (old C/C++ standards, older compilers). The current code in 17u+ is fine with `-std=c++14`.

Incidentally, if we are fixing 11u to compile with GCC 12, should we not also enable a GHA build with GCC 12 in a separate PR so that it does not regress again?

jmtd:
I'll re-label approval-request, a I doubt Goetz saw my comment (or yours @gnu-andrew )

/approval request rationale for change added in https://github.com/openjdk/jdk11u-dev/pull/2470#issuecomment-1927337917 , please take another look!

@gnu-andrew : 
> Incidentally, if we are fixing 11u to compile with GCC 12, should we not also enable a GHA build with GCC 12 in a separate PR so that it does not regress again?

Makes sense to me. I guess that would start with a new 11u-specific JDK bug; I'll file one.

gnu-andrew:
> I'll re-label approval-request, a I doubt Goetz saw my comment (or yours @gnu-andrew )

I think #2451 needs to be re-labelled as well. I'm not a fan of removing these requests outside the process used by the approval/approve commands as it isn't clear what's going on from the PR.

PR #2609, opened 2024-03-18
Backport 6c0bebccb0092d9726eb89a054e023e92edf7ca6

SoniaZaldana:
Hi all, 

This PR contains a backport of [6c0bebccb0092d9726eb89a054e023e92edf7ca6](https://github.com/openjdk/jdk/commit/6c0bebccb0092d9726eb89a054e023e92edf7ca6). 

Patch is not clean. Test won't build in 11 because it can't find ```import jdk.internal.util.OperatingSystem```. 

This class was introduced in 21 as part of [JDK-8303485](https://bugs.openjdk.org/browse/JDK-8303485). 

I've worked around that by removing the dependency to that utility function and checking the operating system with ```System.getProperty("os.name")```. 

Testing: 
- [x] Test passes in linux 
- [ ] Test passes in mac with GHA. 

Thanks!


amosshi:
SAP nightlies passed on `2024-04-16,17,19,20`

RealLucy (approved):
LGTM.

PR #2995, opened 2025-02-14
Backport 955bae2b03055552237584aa7c68b1afcfa3b9c1

RadekCap:
Backport of JDK-8342496 - C2/Shenandoah: SEGV in compiled code when running jcstress

It's fixing the java/lang/invoke/PrivateInvokeTest.java#PrivateInvokeTest failure.

Clean backport. 
Passed tier 1 tests. 
Passed gtests.


jerboaa:
@RadekCap Did you run the test without the patch an with it? If so what were the results?

RadekCap:
No, because in the bug it 11is  listed as an affected version.

RadekCap:
("It's fixing the java/lang/invoke/PrivateInvokeTest.java#PrivateInvokeTest failure." has been removed from the first comment as it was relict from the template which shouldn't be there.)

jerboaa:
> No, because in the bug 11is listed as an affected version.

Please do so. In fact, it's the least one should do to sanity-check a patch. I'm not sure whether this test has been run on JDK 11. Thanks!

PR #3003, opened 2025-02-26
8315380: AsyncGetCallTrace crash in frame::safe_for_sender

jbachorik:
This change is fixing the problem in `frame_aarch64.cpp`, function `safe_for_sender`, where we have this code

```
bool unextended_sp_safe = unextended_sp < thread->stack_base();
```

While this captures one possibility of not being safe, it omits the check for `unextended_sp` falling within the stack space.

The proposed change then is

```
bool unextended_sp_safe = (unextended_sp < thread->stack_base() && \
                             sp >= thread->stack_base() - thread->stack_size());
```

This is actually just making sure the behaviour is the same as in JDK 15+ (since [JDK-8238988](https://bugs.openjdk.org/browse/JDK-8238988)) where the `unextended_sp` is checked for being within the stack limits.

The change is not accompanied by a JTReg test because I was not able to craft one triggering the issue reliably.

Existing tests from tier1-tier4 were run on a linux-aarch64 system with no new failures observed.

---------
### Progress
- [x] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))
- [x] [JDK-8315380](https://bugs.openjdk.org/browse/JDK-8315380) needs maintainer approval

### Issue
 * [JDK-8315380](https://bugs.openjdk.org/browse/JDK-8315380): AsyncGetCallTrace crash in frame::safe_for_sender (**Bug** - P4 - Approved)

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)
 * [Andrei Pangin](https://openjdk.org/census#apangin) (@apangin - no project role) Review applies to [d8e1257f](https://git.openjdk.org/jdk11u-dev/pull/3003/files/d8e1257f3eaa32131561fc24e7cae4667086a805)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/3003/head:pull/3003` \
`$ git checkout pull/3003`

`$ git checkout pull/3003` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/3003/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3003`

`$ git pr show -t 3003`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/3003.diff">https://git.openjdk.org/jdk11u-dev/pull/3003.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/3003#issuecomment-2684671800)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/frame_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/frame_aarch64.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2021, Red Hat Inc. All rights reserved.
+ * Copyrithg (c) 2025, Datadog, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -84,7 +85,8 @@ bool frame::safe_for_sender(JavaThread *thread) {
   // So unextended sp must be within the stack but we need not to check
   // that unextended sp >= sp
 
-  bool unextended_sp_safe = (unextended_sp < thread->stack_base());
+  bool unextended_sp_safe = (unextended_sp < thread->stack_base() && \
+                             sp >= thread->stack_base() - thread->stack_size());
 
   if (!unextended_sp_safe) {
     return false;
--- a/src/hotspot/cpu/aarch64/frame_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/frame_aarch64.cpp
@@ -86,7 +86,7 @@ bool frame::safe_for_sender(JavaThread *thread) {
   // that unextended sp >= sp
 
   bool unextended_sp_safe = (unextended_sp < thread->stack_base() && \
-                             sp >= thread->stack_base() - thread->stack_size());
+                             unextended_sp >= thread->stack_base() - thread->stack_size());
 
   if (!unextended_sp_safe) {
     return false;
--- a/src/hotspot/cpu/aarch64/frame_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/frame_aarch64.cpp
@@ -85,8 +85,7 @@ bool frame::safe_for_sender(JavaThread *thread) {
   // So unextended sp must be within the stack but we need not to check
   // that unextended sp >= sp
 
-  bool unextended_sp_safe = (unextended_sp < thread->stack_base() && \
-                             unextended_sp >= thread->stack_base() - thread->stack_size());
+  bool unextended_sp_safe = (unextended_sp < thread->stack_base() && unextended_sp >= thread->stack_end());
 
   if (!unextended_sp_safe) {
     return false;
--- a/src/hotspot/cpu/aarch64/frame_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/frame_aarch64.cpp
@@ -1,7 +1,6 @@
 /*
  * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2021, Red Hat Inc. All rights reserved.
- * Copyrithg (c) 2025, Datadog, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it


jbachorik:
> Should not the new check be of unextended_sp rather than sp? That would match the check in JDK 17.

Indeed. I feel a bit stupid now. Fixed.

phohensee (approved):
Looks good now.

jbachorik:
/approval I would like to ask for approval to integrate this JDK 11 specific bug fix. The change is very limited and improves the profiling experience for users of this Java version.

jbachorik:
/approval request I would like to ask for approval to integrate this JDK 11 specific bug fix. The change is very limited and improves the profiling experience for users of this Java version.

jerboaa:
@jbachorik This is in a delicate area of the JVM and is a JDK 11u specific fix. Please get a second review for this. Thanks! It would also be good to explain why this isn't an issue in later JDKs on the issue.

jbachorik:
Ok, I bumped up the number of required reviewers. 
But I don't seem to be able to find any mechanism to actually request the second review anywhere.

So, if someone is reading this in the email thread and has a few moments to review this rather trivial change, please, have a look.

src/hotspot/cpu/aarch64/frame_aarch64.cpp
```diff
@@ -84,7 +85,8 @@ bool frame::safe_for_sender(JavaThread *thread) {
   // So unextended sp must be within the stack but we need not to check
   // that unextended sp >= sp
 
-  bool unextended_sp_safe = (unextended_sp < thread->stack_base());
+  bool unextended_sp_safe = (unextended_sp < thread->stack_base() && \
+                             unextended_sp >= thread->stack_base() - thread->stack_size());
```
shipilev, line 89: 1. What's the meaning of trailing `\` here? 
2. I think you can do `unextended_sp >= thread->stack_end()` to better capture the intent and match 8238988 better.

shipilev (commented):
I think @apangin would be good to look at it.

src/hotspot/cpu/aarch64/frame_aarch64.cpp
jbachorik: > What's the meaning of trailing \ here?
No particular meaning, removed.

> I think you can do unextended_sp >= thread->stack_end() to better capture the intent and match 8238988 better.
Yes. Done.

Also, I noticed that the check for unextended_sp is not done very consistently across archs. This is also changed in 8238988 but I opted for the simplicity and added the change only for the arch for which the original issue was reported. I hope it's ok but if a more extensive change is preferred I can apply similar logic to other archs as well.

apangin (approved):

jbachorik:
@jerboaa AFAICS, it is already mentioned in [this comment](https://bugs.openjdk.org/browse/JDK-8315380?focusedId=14754907&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14754907), along with the JBS ticket that made it safe in JDK 15+

src/hotspot/cpu/aarch64/frame_aarch64.cpp
shipilev: I see this for other arches:
 - AArch64: Checks (stack_base, *)
 - ARM: Checks [stack_base, sp]
 - PPC: Checks (stack_base, *)
 - S390: Checks (stack_base, *)
 - SPARC: Checks [stack_base, sp]
 - x86: Checks (stack_base, sp]

So I think only AArch64, PPC and S390 are affected by this bug? Checking against `sp` on other arches looks more conservative than checking for `stack_end()`, so I think we are "fine" there. Given that profiling on PPC and S390 is likely not happening all that often -- I think async-profiler started supporting PPC profiling only not that long ago -- I don't think there is a need to fix those.

@TheRealMDoerr might disagree, though :)

shipilev (approved):
Fine by me, thanks.

jbachorik:
Thanks everyone for helping me to review this!

@jerboaa I think we are ok now?

src/hotspot/cpu/aarch64/frame_aarch64.cpp
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, 2021, Red Hat Inc. All rights reserved.
+ * Copyrithg (c) 2025, Datadog, Inc. All rights reserved.
```
jerboaa, line 4:
```suggestion
 * Copyright (c) 2025, Datadog, Inc. All rights reserved.
```

I tend to agree that this doesn't really need a copyright update.

jbachorik: Ok. Removed. 
Usually I was asked to add a copyright line so I was pro-active this time :)
But I have no strong opinion about this.

phohensee (approved):

phohensee (approved):

shipilev (approved):
Still fine.

jerboaa:
Please integrate this.

PR #3012, opened 2025-03-13
Backport JDK-8351933

dchuyko:
This a backport of JDK-8351933 [0] (PR [1]) for 11u. At one of code paths the TC subfield of ctl field is decremented and the result is not masked correctly. The target code is in runWorker() instead of tryTrim()/awaitWork() and the surrounding code is different from the current master, as well as the original mask name. The core change is the same, candidate for CTL.compareAndSet is constructed using '(RC_MASK & c) | (TC_MASK & (c - TC_UNIT))' instead of '(UC_MASK & (c - TC_UNIT))' to correctly preserve the RC subfield.

[0] https://bugs.openjdk.org/browse/JDK-8351933
[1] https://github.com/openjdk/jdk/pull/24034


DougLea (approved):
Yes, this corrects i the same way as in other releases

PR #3026, opened 2025-04-25
8348597: Update HarfBuzz to 10.4.0

vieiro:
Backport of [JDK-8348597](https://bugs.openjdk.org/browse/JDK-8348597) from JDK17 that updates HarfBuzz to 10.4.0 (which [improves drawing performande by 10+% and includes different build fixes](https://newreleases.io/project/github/harfbuzz/harfbuzz/release/10.4.0)). Backport is not clean because it required adapting `make/lib/Awt2dLibraries.gmk` to fit JDK11.

Even though the backport built correctly, the resulting `jdk/lib/libfontmanager.so` contained undefined symbols that caused Swing tests to fail with `UnsatisfiedLinkError`. This is because JDK11 is missing [JDK-8334964](https://bugs.openjdk.org/browse/JDK-8334964) that excludes hb-subset from compilation, so this is also included as the second commit in the pull request.

While at it, a third commit adds an additional `-Wno-attributes` flag to `HARFBUZZ_DISABLED_WARNINGS_CXX_gcc`, restoring the capability to build JDK11 on Linux with `gcc-4.8.5`  (this is an old version of gcc, dating from 2015, but it's still the system gcc version in RHEL-7).

Tested 

- on Windows 10 with `jdk:jfc_demo`. 
- on Linux with `jdk:jfc_demo`, with 
    - `gcc 4.8.5/harfbuzz-1.7.5 (x86_64 & s390x)`, 
    - `gcc 8.5.0/harfbuzz-1.7.5 (x86_64)`, 
    - `gcc 11.5.0/harfbuzz-2.7.4 (x86_64)` 
    - and `gcc 14.2.1/harfbuzz-9.0.0 (x86_64)`

with both `--with-harfbuzz=system` and `--with-harfbuzz=bundled`.


vieiro:
Fixing issues:
/issue add JDK-8319197
/issue remove JDK-8334964

make/lib/Awt2dLibraries.gmk
```diff
@@ -555,24 +555,40 @@ else
...
     HARFBUZZ_CFLAGS += -DHAVE_SOLARIS_ATOMIC_OPS
   endif
-  # hb-ft.cc is not presently needed, and requires freetype 2.4.2 or later.
-  LIBFONTMANAGER_EXCLUDE_FILES += libharfbuzz/hb-ft.cc
+  
+   # Early re-canonizing has to be disabled to workaround an internal XlC compiler error
+   # when building libharfbuzz
+   ifeq ($(call isTargetOs, aix), true)
+     HARFBUZZ_CFLAGS += -qdebug=necan
+   endif
```
jerboaa, line 563: This is actually a change from [JDK-8258484](https://bugs.openjdk.org/browse/JDK-8258484) which is not in JDK 11u. Intentional? If yes, we should probably use `/issue add JDK-8258484` to have it tracked.

make/lib/Awt2dLibraries.gmk
```diff
@@ -555,24 +555,40 @@ else
...
-       undef missing-field-initializers deprecated-declarations c++11-narrowing range-loop-analysis
-  HARFBUZZ_DISABLED_WARNINGS_microsoft := 4267 4244 4090 4146 4334 4819 4101 4068 4805 4138
+   HARFBUZZ_DISABLED_WARNINGS_gcc := type-limits missing-field-initializers strict-aliasing \
+        array-bounds parentheses dangling-pointer
+   # noexcept-type required for GCC 7 builds. Not required for GCC 8+.
+   # expansion-to-defined required for GCC 9 builds. Not required for GCC 10+.
+   # maybe-uninitialized required for GCC 8 builds. Not required for GCC 9+.
+   # calloc-transposed-args required for GCC 14 builds. (fixed upstream in Harfbuzz 032c931e1c0cfb20f18e5acb8ba005775242bd92)
+   HARFBUZZ_DISABLED_WARNINGS_CXX_gcc := reorder delete-non-virtual-dtor strict-overflow \
+        maybe-uninitialized class-memaccess unused-result extra use-after-free noexcept-type \
```
jerboaa, line 583: This also adds `use-after-free` as a disabled warning for `gcc`. That is part of [JDK-8286562](https://bugs.openjdk.org/browse/JDK-8286562) which we probably don't want to partially introduce here.

make/lib/Awt2dLibraries.gmk
```diff
@@ -555,24 +555,40 @@ else
...
-  # noexcept-type required for GCC 7 builds. Not required for GCC 8+.
-  # expansion-to-defined required for GCC 9 builds. Not required for GCC 10+.
-  HARFBUZZ_DISABLED_WARNINGS_CXX_gcc := reorder delete-non-virtual-dtor strict-overflow \
-       maybe-uninitialized class-memaccess unused-result extra noexcept-type expansion-to-defined
-  HARFBUZZ_DISABLED_WARNINGS_clang := unused-value incompatible-pointer-types \
-       tautological-constant-out-of-range-compare int-to-pointer-cast \
-       undef missing-field-initializers deprecated-declarations c++11-narrowing range-loop-analysis
-  HARFBUZZ_DISABLED_WARNINGS_microsoft := 4267 4244 4090 4146 4334 4819 4101 4068 4805 4138
+   HARFBUZZ_DISABLED_WARNINGS_gcc := type-limits missing-field-initializers strict-aliasing \
+        array-bounds parentheses dangling-pointer
```
jerboaa, line 577: `parentheses dangling-pointer` are new here. `danling-pointer` seems to be part of the JDK 17 harfbuzz update to 10.4.0. `parenthesis` not. Is it needed?

jerboaa (commented):
A few comments that need clarifications. Thanks!

make/lib/Awt2dLibraries.gmk
vieiro: Yep. this is intentional. Will add the issue.

make/lib/Awt2dLibraries.gmk
vieiro: Not strictly required (gcc 14, gcc 4.8.5). Will remove, thanks!

make/lib/Awt2dLibraries.gmk
vieiro: Not strictly required (gcc 14, gcc 4.8.5). Will remove, thanks!

jerboaa (approved):

gnu-andrew:
Will take a look at this.
/reviewers 2 reviewer

vieiro:
While at it, let's add MacOS 12.6.1/arm64 (virtualized)/clang 13.1.6 to the list of tests.

make/lib/Awt2dLibraries.gmk
gnu-andrew: Why are you including this? Have you built this on AIX?

gnu-andrew (changes requested):
The code changes look fine and match 17u.

I'm confused by what is going on in `make/lib/Awt2dLibraries.gmk`.  There should be just two changes here according to the description:

* The addition of `dangling-pointer` to the disabled warnings as in 8348597
* The expansion of `LIBFONTMANAGER_EXCLUDE_FILES` from 8319197

While I would prefer 8319197 was handled separately - as it is independent of this change - it is simple enough that it can be included here.

What I don't get is why the block is being indented by one space, which is obscuring the changes made to the set of disabled warnings. I think the clang ones are also being reordered but it's hard to tell for sure. I imagine this is why the other additions Severin already caught slipped in as well.

I also don't think an AIX change should be included if you are not testing on AIX or there is at least some relation to this change.

Please fix `Awt2dLibraries.gmk` to just update the lines we actually want to change as documented in the PR summary without altering the indentation and adding the AIX fix.  Thanks.

vieiro:
Thanks for the review!

> I'm confused by what is going on in `make/lib/Awt2dLibraries.gmk`. There should be just two changes here according to the description:
> 
>     * The addition of `dangling-pointer` to the disabled warnings as in 8348597
> 
>     * The expansion of `LIBFONTMANAGER_EXCLUDE_FILES` from 8319197
> 
> 

The intention was to make `Awt2dLibraries.gmk` more similar to the one in JDK-17 (which has had significant changes during the years), with the intention to make future backports easier (at the cost of a more difficult review), hence the reordering and the indentation.

So I've re-arranged the changes to make them more similar to the original `Awt2dLibraries.gmk` in 11. Hope this makes it easier to review. Note the `attributes` is required for `gcc.4.8.5`.

Retested with:

-   gcc 4.8.5/harfbuzz-1.7.5 (x86_64),
-   and gcc 14.2.1/harfbuzz-9.0.0 (x86_64)

jerboaa:
This needs to conclude review before approval. Removing labels meanwhile.

vieiro:
Hi. Sorry for the delay.

Here's a list of some of the compiler flags and the commits in 17 where they were introduced. This may help us see what we should include here and what may want to backport later.

- `dangling-pointer stringop-overflow` in `DISABLED_WARNINGS_gcc`:

This was introduced in 2023 as [8300692: GCC 12 reports some compiler warnings in bundled freetype](https://github.com/openjdk/jdk17u-dev/commit/5fd778be5a40a45523d9200c866a73ddb577bd56).

- `danglling-reference` in `HARFBUZZ_DISABLED_WARNINGS_CXX_gcc`:

This was introduced in 2023 as [8308283: Build failure with GCC12 & GCC13](https://github.com/openjdk/jdk17u-dev/commit/e3dad43a7dda8be42caf5ccaed4dfe0d51a52cf5)

- `calloc-transposed-args` in  `HARFBUZZ_DISABLED_WARNINGS_CXX_gcc`:

This one from 2024, introduced in [8324243: Compilation failures in java.desktop module with gcc 14](https://github.com/openjdk/jdk17u-dev/commit/0e524e2f476835885b34e221e64d8e12a9a1d792)

- `attributes` We're introducing it here to be able to build on `rhel7` with the system bundled `gcc-4.8`. Otherwise the build fails with the new harfbuzz code, with errors similar to:
```
./src/java.desktop/share/native/libharfbuzz/hb-serialize.hh:627:59: error: 'returns_nonnull' attribute directive ignored [-Werror=attributes]
   Type *start_embed (const Type *obj HB_UNUSED = nullptr) const
```

I think I'll trim out as many flags of these flags as possible in this commit, just to be able to build with `gcc8` on `rhel7` and with `gcc10` (GHA). We may want to backport the stuff above to be able to build on more platforms.

vieiro:
So this drops the flags required for:
- building with GCC 4.x (requires `attributes` for building the new harfbuzz bits introduced here)
- gcc 12 and up (would require backporting 8300692, 8308283, 8324243).

This builds correctly with GCC 8.X (rhel-7, rhel-8) and in GCC 10.X (GHA). Let me know if you want to add any of these back again. 

Thanks for the review!

vieiro:
@gnu-andrew  Restored!

Yes, let's keep 8300692, 8308283, 8324243 in mind in case somebody needs building with gcc > 12.

vieiro:
/approval request Please consider approving this backport from JDK17 that levels up HarfBuzz version with upper JDKs and improves drawing performance.

vieiro:
> @vieiro please re-request approval for this.

Note there's a missing review, though.

jerboaa:
I'll review this again today.

jerboaa (approved):
OK.

vieiro:
Thanks for the reviews and approval!

/integrate

PR #3074, opened 2025-08-22
8298783: java/lang/ref/FinalizerHistogramTest.java failed with "RuntimeException: MyObject is not found in test output"

sendaoYan:
Hi all,

FinalizerHistogramTest exercises the internal java.lang.ref.FinalizerHistogram API. The test:

1. Creates 1000 MyObject instances, each with a finalize() that acquires a ReentrantLock held by the main thread, permanently blocking the finalizer thread on the first finalized object.

2. Triggers GC and waits until reference/finalizer processing has progressed.

3. Reads FinalizerHistogram.getFinalizerHistogram() and expects to see MyObject entries — objects still waiting in the finalizer queue while the finalizer thread is blocked.

The intermittent failure (RuntimeException: MyObject is not found in test output) was a test bug, not a product bug. The old test called System.gc() and then spun on while (wasTrapped < 1), proceeding as soon as the first finalize() was entered. At that point, other MyObject instances might not yet have been discovered and enqueued for finalization, so the histogram could legitimately be empty. 

This backport makes the test more robust by:

- Running in othervm, since the test blocks the global finalizer thread.

- Replacing non-atomic volatile int counters with AtomicInteger.

- Using WhiteBox.fullGC() plus a loop on WhiteBox.waitForReferenceProcessing() instead of System.gc() plus a busy-wait on wasTrapped.

- Adapting the WhiteBox path for jdk11u (sun.hotspot.WhiteBox / ClassFileInstaller instead of jdk.test.whitebox.WhiteBox).

This PR is intend to fix the test bug which cause test intermittent fails. Backport not clean because the path of `sun.hotspot.WhiteBox` different to main-line, the path should adopt to jdk11u-dev when backport from main-line to jdk11u-dev, Other parts is cleanly.

I have verified this backport plus with https://github.com/openjdk/jdk11u-dev/pull/3105.

Thanks!

---------

---------
### Progress
- \[x\] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- \[ \] [JDK-8298783](https://bugs.openjdk.org/browse/JDK-8298783) needs maintainer approval

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Issue
 * [JDK-8298783](https://bugs.openjdk.org/browse/JDK-8298783): java/lang/ref/FinalizerHistogramTest.java failed with "RuntimeException: MyObject is not found in test output" (**Bug** - P4)

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/3074/head:pull/3074` \
`$ git checkout pull/3074`

`$ git checkout pull/3074` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/3074/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3074`

`$ git pr show -t 3074`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/3074.diff">https://git.openjdk.org/jdk11u-dev/pull/3074.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/3074#issuecomment-3212872857)
</details>


sendaoYan:
/approval request Clean backport to make test more robustness which fix the test fails intermittent. Change has been verifiled locally on linux-x64. Test-fix only, risk is low.

sendaoYan:
I think we should backport JDK-8305186 first.

sendaoYan:
The patch different to mainline:

```diff
diff --git a/test/jdk/java/lang/ref/FinalizerHistogramTest.java b/test/jdk/java/lang/ref/FinalizerHistogramTest.java
index 192c229c7b5..4dcea55b1f4 100644
--- a/test/jdk/java/lang/ref/FinalizerHistogramTest.java
+++ b/test/jdk/java/lang/ref/FinalizerHistogramTest.java
@@ -21,7 +21,7 @@
  * questions.
  */
 
-import jdk.test.whitebox.WhiteBox;
+import sun.hotspot.WhiteBox;
 
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
@@ -34,8 +34,8 @@ import java.lang.reflect.Field;
  * @summary Unit test for FinalizerHistogram
  * @modules java.base/java.lang.ref:open
  * @library /test/lib
- * @build jdk.test.whitebox.WhiteBox
- * @run driver jdk.test.lib.helpers.ClassFileInstaller jdk.test.whitebox.WhiteBox
+ * @build sun.hotspot.WhiteBox
+ * @run driver ClassFileInstaller sun.hotspot.WhiteBox
  * @run main/othervm
  *      -Xbootclasspath/a:.
  *      -XX:+UnlockDiagnosticVMOptions
```

phohensee (approved):

sendaoYan:
Waiting https://github.com/openjdk/jdk11u-dev/pull/3105 merged.

sendaoYan:
/approval request Clean backport to make test more robustness which fix the test fails intermittent. Change has been verifiled locally on linux-x64. Backport parity with 11.0.31-oracle, test-fix only, risk is low.

gnu-andrew:
Not sure what is going on with oca as I've seen patches from you before.

tstuefe:
Can you please describe what the patch does? Even though it seems simple enough. In general, it saves us time when reviewing, especially in cases like these where the JBS issue does not explain what went wrong or what the fix does (the issue text only lists symptoms, and the comments don't explain anything either).

Even if the patch is simple and self-explanatory, this helps gauge the importance of issues, whether the author has understood the issue and also whether the author's explanation matches the actual patch. For intermittent errors, it would also help to know how intermittent intermittent is (if this is known) with respect to the old stable target release.

test/jdk/java/lang/ref/FinalizerHistogramTest.java
```diff
@@ -32,37 +33,49 @@
...
-            for(int i = 0; i < objectsCount; ++i) {
+            for(int i = 0; i < OBJECTS_COUNT; ++i) {
                 new MyObject();
             }
-            System.out.println("Objects intialized: " + objectsCount);
-            System.gc();
-            while(wasTrapped < 1);
+            System.out.println("Objects intialized: " + initializedCount.get());
+            wb = WhiteBox.getWhiteBox();
+            wb.fullGC();
```
tstuefe, line 73: Why the change from System.gc to WB.fullGC?

sendaoYan: Use WB.fullGC instead of System.gc is a good practice sometimes, this will make test more robustness and stable, such as https://github.com/openjdk/jdk/pull/29729.

System.gc() is only a hint to the JVM — it does not guarantee that a full collection runs, that all unreachable objects are collected, or that reference/finalizer processing has completed. In this test that was the root cause of the intermittent failure: after System.gc(), the old code spun on while (wasTrapped < 1), i.e. it proceeded as soon as the first MyObject.finalize() was entered. At that moment other MyObject instances might not yet have been discovered and enqueued for finalization, so FinalizerHistogram could legitimately show no MyObject entries.

WhiteBox.fullGC() (see WB_FullGC in whitebox.cpp) forces a full heap collection with GCCause::_wb_full_gc and clears soft references, so unreachable MyObject instances are actually collected. The follow-up loop on wb.waitForReferenceProcessing() (which delegates to Reference.waitForReferenceProcessing()) then waits until pending reference processing — including work handed off to the finalizer thread — has quiesced (false means no active/pending processing). Only after that do we read the histogram, which matches the test's intent: at least one object blocked in finalize(), with others still waiting in the finalizer queue.

So the change is not “use WhiteBox for fun”; it replaces a racy System.gc() + busy-wait with deterministic GC + reference-processing synchronization, which is the actual fix for JDK-8298783.

sendaoYan:
> Can you please describe what the patch does? Even though it seems simple enough. In general, it saves us time when reviewing, especially in cases like these where the JBS issue does not explain what went wrong or what the fix does (the issue text only lists symptoms, and the comments don't explain anything either).
> 
> Even if the patch is simple and self-explanatory, this helps gauge the importance of issues, whether the author has understood the issue and also whether the author's explanation matches the actual patch. For intermittent errors, it would also help to know how intermittent intermittent is (if this is known) with respect to the old stable target release.

Sorry, I have updated the PR description.

I can not reproduce the intermittent failure locally. But the same failure was observed by jdk11u-dev GHA(Seen the comment in https://bugs.openjdk.org/browse/JDK-8298783). This is test-fix only, I think the risk is very low.

PR #3077, opened 2025-08-26
8366125: [11u] Test compiler/loopopts/TestRangeCheckPredicatesControl.java fails OOM

sendaoYan:
Hi all,

On ours linux-x64 machine, `-Xmx256M -XX:+UnlockExperimentalVMOptions -XX:+UseZGC` can not start JVM because of OOM. I think it's not a JVM bug in jdk11u, I think we should increase the max heap size to make test run normally after [JDK-8251949](https://bugs.openjdk.org/browse/JDK-8251949). The `-Xmx256m` was introduced by [JDK-8251949](https://bugs.openjdk.org/browse/JDK-8251949) only for jdk11u. So this PR will only for jdk11u also.

Change has been verified locally, test-fix only, ro risk.

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/TestRangeCheckPredicatesControl.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestRangeCheckPredicatesControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2025, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


sendaoYan:
Looking for reviewer for this test bug fix.

sendaoYan:
GHA report 29 jdi tests fails, the test log shows 'getaddrinfo: unknown host' cause jdi tests fails, this seems like environmental issue. it's unrelated to this PR.

/approval request Increase max heap size for ZGC to avoid OOM during JVM starting. Test-fix, no risk.

jerboaa (approved):

sendaoYan:
Thanks for the reviews and approved.

/integrate

PR #3171, opened 2026-03-11
8310233: Fix THP detection on Linux

tabata-d:
JDK-8310233 is a bug where the JDK 11 Hotspot VM on Linux incorrectly detects Transparent Huge Pages (THPs) status and page size. This leads to misconfigurations and inefficient memory usage.

The fix was originally implemented in JDK 22. Now we are backporting this into JDK 11.
I confirmed that it was reproduced on Linux x86_64 using JDK 11.

## Unclean Backport

- The backport was not entirely clean due to a fundamental dependency on [JDK-8256155](https://bugs.openjdk.org/browse/JDK-8256155), an Enhancement with CSR present in newer JDK versions (e.g., JDK 17+, JDK 22) but absent in JDK 11. A full backport of [JDK-8256155](https://bugs.openjdk.org/browse/JDK-8256155) was avoided as it would introduce significant behavioral changes and compatibility risks unsuitable for JDK 11 in maintenance mode.
- Instead, a targeted manual adaptation was implemented for JDK 11's static huge page mode.
  - **JDK 22's `os::large_page_init()` (with JDK-8256155)**: The JDK 22 code (in the `else` block for static huge pages) explicitly retrieves `os::PageSizes all_large_pages = HugePages::static_info().pagesizes();` . This `all_large_pages` (No longer used in JDK11) is a variable introduced in JDK 22 as part of JDK-8256155 , holds a comprehensive list of all large_page_sizes supported by the kernel not just the default. This allowed JDK 22 to intelligently select the most appropriate large page size (the largest available that is smaller than the bytes being reserved) based on `LargePageSizeInBytes` acting as a _maximum allowed size_. It then uses a complex conditional block to interpret `LargePageSizeInBytes` against this list and populates `_page_sizes` with multiple large page sizes less than or equal to `_large_page_size`.
  - **JDK 11's `os::large_page_init()` (without JDK-8256155)**: The JDK 11 backport's `else` block for static huge pages is significantly simpler. It directly sets `_large_page_size` to `HugePages::default_static_hugepage_size()`. The `all_large_pages` variable and the complex logic used in JDK22 that leverages it for flexible page size selection are not present and not used in the JDK 11 backport. Instead, it populates `_page_sizes` only with the single configured static huge page size and the VM page size.

**Other Code Differences and Adaptations:**
- **`os_linux.cpp`:**
  - The removal of `size_t os::Linux::_default_large_page_size = 0;` (present in JDK 22) was not performed in JDK 11 as this member was absent.
[...]

PR #3210, opened 2026-05-29
8312182: THPs cause huge RSS due to thread start timing issue

tabata-d:
JDK-8312182 is a bug fix that addresses significant Resident Set Size (RSS) bloat in Java applications when Transparent Huge Pages (THP) are enabled on Linux.

The fix was originally implemented in JDK 21. Now we are backporting this into JDK 11.

## Unclean Backport
- The backport was not entirely clean for the C++ source files (`globals_linux.hpp`, `os_linux.cpp`) due to the evolution of internal HotSpot VM APIs and syntax between JDK 11 and JDK 21. 
- **Specific Unclean Parts:**
    1.  **Flag Declaration:** In JDK 21, the `product` macro supports a `DIAGNOSTIC` attribute. This attribute was introduced by JDK-8243208, which is not present in JDK 11. Directly using the JDK 21 syntax would cause a compilation error in JDK 11. For `DisableTHPStackMitigation`, instead of attempting to force the `DIAGNOSTIC` attribute into the `product` macro, it was declared using the `diagnostic` macro directly in `globals_linux.hpp`. This achieves the same intent of marking the flag as diagnostic without requiring the backport of JDK-8243208, which has a broad impact and would introduce significant complexity and risk to the JDK 11 codebase.
    2.  **Enum Access:** Between JDK 11 and JDK 21, the way `THPMode` enum members are accessed changed. In JDK 21, they are accessed using the scope resolution operator (`THPMode::always`), while in JDK 11, they are accessed directly (`THPMode_always`). The code was adjusted to use the direct access method (`THPMode_always`) for enum members in `os_linux.cpp` to align with JDK 11's syntax.
    3.  **Macro Signature:** The `FLAG_SET_ERGO` macro in JDK 21 no longer requires the type (`bool`) as its first argument, but JDK 11 explicitly requires it. The `FLAG_SET_ERGO` macro call in `os_linux.cpp` was modified to explicitly include the `bool` type argument, conforming to JDK 11's macro definition.

## Testing
**System:** Ran on Red Hat Enterprise Linux 9.4 (x86_64).
**jtreg**: A comprehensive `jtreg` run on the entire `hotspot/jtreg` test suite confirmed that **all HotSpot tests passed**. 
**New Test:** The newly added test `THPsInThreadStackPreventionTest.java` passed successfully in both enabled and disabled configurations, confirming the effectiveness of the fix.

---------

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [ ] [JDK-8310687](https://bugs.openjdk.org/browse/JDK-8310687) needs maintainer approval
- [ ] [JDK-8312182](https://bugs.openjdk.org/browse/JDK-8312182) needs maintainer approval

### Integration blocker
&nbsp;⚠️ Dependency #3171 must be integrated first

### Issues
 * [JDK-8312182](https://bugs.openjdk.org/browse/JDK-8312182): THPs cause huge RSS due to thread start timing issue (**Bug** - P3)
 * [JDK-8310687](https://bugs.openjdk.org/browse/JDK-8310687): JDK-8303215 is incomplete (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/3210/head:pull/3210` \
`$ git checkout pull/3210`

`$ git checkout pull/3210` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/3210/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3210`

`$ git pr show -t 3210`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/3210.diff">https://git.openjdk.org/jdk11u-dev/pull/3210.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/3210#issuecomment-4572792843)
</details>

### DIFF
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -82,7 +82,17 @@
           "be dumped into the corefile.")                               \
                                                                         \
   diagnostic(bool, UseCpuAllocPath, false,                              \
-             "Use CPU_ALLOC code path in os::active_processor_count ")
+             "Use CPU_ALLOC code path in os::active_processor_count ")  \
+                                                                        \
+  diagnostic(bool, DisableTHPStackMitigation, false,                    \
+          "If THPs are unconditionally enabled on the system (mode "    \
+          "\"always\"), the JVM will prevent THP from forming in "      \
+          "thread stacks. This switch disables that mitigation and "    \
+          "allows THPs to form in thread stacks.")                      \
+                                                                        \
+  develop(bool, DelayThreadStartALot, false,                            \
+          "Artificially delay thread starts randomly for testing.")     \
+                                                                        \
 
 //
 // Defines Linux-specific default values. The flags are available on all
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -823,6 +823,10 @@ static void *thread_native_entry(Thread *thread) {
     }
   }
 
+  if (DelayThreadStartALot) {
+    os::naked_short_sleep(100);
+  }
+
   // call one more level start routine
   thread->call_run();
 
@@ -962,6 +966,7 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   // Calculate stack size if it's not specified by caller.
   size_t stack_size = os::Posix::get_initial_stack_size(thr_type, req_stack_size);
   size_t guard_size = os::Linux::default_guard_size(thr_type);
+
   // Configure glibc guard page. Must happen before calling
   // get_static_tls_area_size(), which uses the guard_size.
   pthread_attr_setguardsize(&attr, guard_size);
@@ -982,13 +987,16 @@ bool os::create_thread(Thread* thread, ThreadType thr_type,
   }
   assert(is_aligned(stack_size, os::vm_page_size()), "stack_size not aligned");
 
-  // Add an additional page to the stack size to reduce its chances of getting large page aligned
-  // so that the stack does not get backed by a transparent huge page.
-  size_t default_large_page_size = HugePages::default_static_hugepage_size();
-  if (default_large_page_size != 0 &&
-      stack_size >= default_large_page_size &&
-      is_aligned(stack_size, default_large_page_size)) {
-    stack_size += os::vm_page_size();
+  if (!DisableTHPStackMitigation) {
+    // In addition to the glibc guard page that prevents inter-thread-stack hugepage
+    // coalescing (see comment in os::Linux::default_guard_size()), we also make
+    // sure the stack size itself is not huge-page-size aligned; that makes it much
+    // more likely for thread stack boundaries to be unaligned as well and hence
+    // protects thread stacks from being targeted by khugepaged.
+    if (HugePages::thp_pagesize() > 0 &&
+        is_aligned(stack_size, HugePages::thp_pagesize())) {
+      stack_size += os::vm_page_size();
+    }
   }
 
   int status = pthread_attr_setstacksize(&attr, stack_size);
@@ -3502,6 +3510,27 @@ bool os::Linux::libnuma_init() {
 }
 
 size_t os::Linux::default_guard_size(os::ThreadType thr_type) {
+
+  if (!DisableTHPStackMitigation) {
+    // If THPs are unconditionally enabled, the following scenario can lead to huge RSS
+    // - parent thread spawns, in quick succession, multiple child threads
+    // - child threads are slow to start
+    // - thread stacks of future child threads are adjacent and get merged into one large VMA
+    //   by the kernel, and subsequently transformed into huge pages by khugepaged
+    // - child threads come up, place JVM guard pages, thus splinter the large VMA, splinter
+    //   the huge pages into many (still paged-in) small pages.
+    // The result of that sequence are thread stacks that are fully paged-in even though the
+    // threads did not even start yet.
+    // We prevent that by letting the glibc allocate a guard page, which causes a VMA with different
+    // permission bits to separate two ajacent thread stacks and therefore prevent merging stacks
+    // into one VMA.
+    //
+    // Yes, this means we have two guard sections - the glibc and the JVM one - per thread. But the
+    // cost for that one extra protected page is dwarfed from a large win in performance and memory
+    // that avoiding interference by khugepaged buys us.
+    return os::vm_page_size();
+  }
+
   // Creating guard page is very expensive. Java thread has HotSpot
   // guard pages, only enable glibc guard page for non-Java threads.
   // (Remember: compiler thread is a Java thread, too!)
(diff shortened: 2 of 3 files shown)


tstuefe:
Hi @tabata-d ,

First off, thanks for preparing the patches. I know that is a lot of work. To save you the work and frustration of getting something work-intensive rejected later, it is always good to discuss larger backport projects upfront in the OpenJDK mailing lists first. Not sure which mailing list would be best for that @jerboaa ?

In this case, I am not yet convinced that these backports are necessary, at least not in their full form. Both @jerboaa and @gnu-andrew have reservations, and they are the maintainers who need convincing.

---------------------------

JDK-8312182 bypasses a problem caused by the behavior of Java, kernel, and glibc:

- java orders glibc to create thread stacks without glibc guard pages since we create our own guard pages.
- glibc mmap the memory for thread stack
- on the system level, we use THP mode "always"
- (concurrently) khugepaged decides to coalesce the start of the stacks into huge pages. If multiple stacks are directly adjacent, from the kernel's perspective they are one VMA, and it aggressively converts that VMA into huge pages, making the memory resident as a result
- Java then places its stack guard, which splinters the VMA again, but the damage is already done - all that memory is resident, and thus RSS goes up.

But that was then. Things are different now, since java sure was not the only program suffering from THP RSS bloat. Therefore:

1) glibc has a tunable since glibc 2.38, called **glibc.pthread.stack_hugetlb**. That is by default 1, but can be set to 0, and that causes glibc to mmap the thread stacks with *MADV_NOHUGEPAGE*, which should completely prevent the problem.
2) Kernels are a lot smarter now about when to do MMAP coalescing. It will be a lot less likely to promote VMAs to THP that are sparsely populated, and stacks of freshly created stacks are just that.

So we have the following workarounds:
- on new kernels (starting at around 6.14... ??) nothing should be needed at all 
- on new glibcs >= 2.38 we can use the tunable
- obviously, we can switch THP mode to "madvise", which affects the whole machine. But tbh on older kernels where the THP daemon is that aggressive this may be good advise anyway. Because the problem does not only affect java stacks.
- Yet another obvious workaround is to use stack sizes just a wee bit smaller than THP size, so not 2MB but e.g. 1.9MB. 

@tabata-d where, exactly, can you reproduce the problem? Which OS/kernel version/glibc release?

jerboaa:
> Not sure which mailing list would be best for that @jerboaa ?

`jdk8u-dev@openjdk.org` for OpenJDK 8u backports, `jdk-updates-dev@openjdk.org` for any other (newer) release.

tabata-d:
Hi @tstuefe,

Thank you for the detail explanations. I now understand that this is not a fatal issue for newer glibc/kernel, because there are useful workarounds available.

> @tabata-d where, exactly, can you reproduce the problem? Which OS/kernel version/glibc release?

The issue can be reproduced in the following environments:

1. RHEL 8.10
kernel: 4.18.0
glibc: 2.28

2. RHEL 9.6
kernel: 5.14.0
glibc: 2.34

It can be reproduced on RHEL 8 and 9 series, which still have a large number of users.
On RHEL 10 (glibc 2.39), it might be possible to workaround it using **glibc.pthread.stack_hugetlb**, but I haven’t tried it yet.

tstuefe:
@tabata-d Yes, I did my own tests today and saw the bug (very clearly) still on RHEL8 but not that clearly on RHEL9.

On RHEL 10 it should be pretty much gone (I cannot reproduce the problem on kernels >= 6), since newer Khugepaged versions are a lot smarter about coalescing large sparse memory areas to huge pages. Hugepaged on older releases is way too aggressive.

If you can, can you share some more details about the RHEL 9 tests?
- what architecture (arm64 or x64?)
- how many cores has the box
- what are the khugepaged settings (basically everything inside `/sys/kernel/mm/transparent_hugepage/khugepaged/`?
- How exactly did you reproduce the bug? What are the RSS numbers you see?

Sorry, we can't really take the full chain of backports you did - just too risky and invasive. I would ask that you close your THP-related PRs - I know it feels bad to have spent time on something that gets rejected, but we need to keep the risk minimal.

I will fix this problem in JDK 11 and 8 with a small, targeted fix specific to these older releases. Your work helped to make us aware that this was a real problem, thanks for that. The problem will be fixed for your users (probably not for the July update, but for the October update).

tabata-d:
@tstuefe Sorry for the late reply.

> If you can, can you share some more details about the RHEL 9 tests?

OS: RHEL 9.6
Architecture: x64
Machine cores: 4 cores

khugepaged settings:
```
$ find /sys/kernel/mm/transparent_hugepage/khugepaged/ -type f | while read f; do echo "$f"; cat
"$f"; done
/sys/kernel/mm/transparent_hugepage/khugepaged/defrag
1
/sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_shared
256
/sys/kernel/mm/transparent_hugepage/khugepaged/scan_sleep_millisecs
10000
/sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none
511
/sys/kernel/mm/transparent_hugepage/khugepaged/pages_to_scan
4096
/sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_swap
64
/sys/kernel/mm/transparent_hugepage/khugepaged/alloc_sleep_millisecs
60000
/sys/kernel/mm/transparent_hugepage/khugepaged/pages_collapsed
69
/sys/kernel/mm/transparent_hugepage/khugepaged/full_scans
1
```

Reproducer program:
[Test.java](https://github.com/user-attachments/files/28690694/Test.java)

Test.java prints `/proc/self/status` at runtime.

When I ran Test.java in the above environment, I got the following result:
```
$ ./jdk-11.0.31+11/bin/java -Xss2m -XX:+AlwaysPreTouch Test
Process PID: 100323

--- /proc/self/status ---
VmSize: 28563792 kB
VmRSS:  13053880 kB
Threads:        10020
-------------------------
```

After changing the THP configuration from `always` to `madvise`, I got the following result:
```
$ ./jdk-11.0.31+11/bin/java -Xss2m -XX:+AlwaysPreTouch Test
Process PID: 12219

--- /proc/self/status ---
VmSize: 28566808 kB
VmRSS:    693716 kB
Threads:        10020
-------------------------
```

On my RHEL 9 environment, I observed a significant difference in RSS depending on the THP configuration.

tabata-d:
> Sorry, we can't really take the full chain of backports you did - just too risky and invasive. I would ask that you close your THP-related PRs - I know it feels bad to have spent time on something that gets rejected, but we need to keep the risk minimal.

I will close the series of PR.
I was reminded that JDK 11 and 8 are very stable releases, and that unclean and large backports like this one are considered too risky to accept. From now on, I will consult the mailing list before working on such changes.

@tstuefe @jerboaa @gnu-andrew 
Thank you very much for the courteous responses and also for your continuous maintenace work.

PR #3203, opened 2026-05-21
8384815: SelectOneKeyOutOfMany and PreferredKey fail after expired test certificate

sendaoYan:
Hi all,

This is backport PR to make test generate certificates during test run which avoid the fixed certificate expired cause test fails.
There are two changes make this PR backported uncleanly.

1. jdk11u lack of JDK-8325766 cause javac report can not find `setOneHourValidity`. I add setOneHourValidity() manully.
2. Class CertificateBuilder locate in different package in jdk17u and jdk11u. So I adopt the @libary and import package name for CertificateBuilder.
3. CertificateBuilder.setPublicKey() return void in jdk11u, so split the chaind calls for new CertificateBuilder()
4. jdk11u do not support Arrays.stream().toList(), so I use Arrays.stream.anyMatch() instead.

Change has been verified locally by run the releated on linux-x64. Test-fix only, no risk.

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[ \] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[ \] [JDK-8384815](https://bugs.openjdk.org/browse/JDK-8384815) needs maintainer approval

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Issue
 * [JDK-8384815](https://bugs.openjdk.org/browse/JDK-8384815): SelectOneKeyOutOfMany and PreferredKey fail after expired test certificate (**Bug** - P2)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/3203/head:pull/3203` \
`$ git checkout pull/3203`

Update a local copy of the PR: \
`$ git checkout pull/3203` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/3203/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 3203`

View PR using the GUI difftool: \
`$ git pr show -t 3203`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #3213, opened 2026-06-01
8244336: Restrict algorithms at JCE layer

TheMangovnik:
Backport of [JDK-8244336](https://bugs.openjdk.org/browse/JDK-8244336) - Restrict algorithms at JCE layer.

This is in preparation for a backport of [JDK-8375549](https://bugs.openjdk.org/browse/JDK-8375549).

This backport will be followed up by [JDK-8367583](https://bugs.openjdk.org/browse/JDK-8367583) and [JDK-8373690](https://bugs.openjdk.org/browse/JDK-8373690) to fix introduced bugs by this backport.

## Extra changes

`src/java.base/share/classes/sun/security/util/CryptoAlgorithmConstraints.java`

On line 61:
```diff
- debug.println("CryptoAlgoConstraints: ", msg);
+ debug.println("CryptoAlgoConstraints: " + msg);
```

I made this change, because otherwise the compilation would not pass complaining about incorrect static method calling.

## Tests

Tests were run on Fedora 44.

### New tests - MIXED

```
==============================
Test summary
==============================
   TEST                                              TOTAL  PASS  FAIL ERROR
   jtreg:test/jdk/javax/crypto/Cipher/TestDisabledAlgorithms.java
                                                         1     1     0     0
	jtreg:test/jdk/java/security/KeyStore/TestDisabledAlgorithms.java
>>                                                       1     0     1     0 <<
	jtreg:test/jdk/java/security/MessageDigest/TestDisabledAlgorithms.java
                                                         1     1     0     0
    jtreg:test/jdk/java/security/Signature/TestDisabledAlgorithms.java
                                                         1     1     0     0
    jtreg:test/jdk/sun/security/pkcs11/Cipher/TestDisabledAlgorithms.java
                                                         1     1     0     0
    jtreg:test/jdk/sun/security/pkcs11/Signature/TestNONEwithRSA.java
                                                         1     1     0     0
    jtreg:test/jdk/sun/security/util/AlgorithmConstraints/InvalidCryptoDisabledAlgos.java
                                                         1     1     0     0
==============================
```

The failing test `test/jdk/java/security/KeyStore/TestDisabledAlgorithms.java` will be fixed in follow up backport of [JDK-8373690](https://bugs.openjdk.org/browse/JDK-8373690).

### `sun/security` - MIXXED

```
==============================
Test summary
==============================
   TEST                                              TOTAL  PASS  FAIL ERROR
>> jtreg:test/jdk/sun/security                         677   676     1     0 <<
==============================
TEST FAILURE
```

The failing test is `sun/security/ssl/X509KeyManager/PreferredKey.java`, afaik this unrelated to this backport and is being addressed with another [PR](https://github.com/openjdk/jdk11u-dev/pull/3203).

### Tier 1 - PASSED

```
==============================
Test summary
==============================
   TEST                                              TOTAL  PASS  FAIL ERROR
   jtreg:test/hotspot/jtreg:tier1                     1497  1497     0     0
   jtreg:test/jdk:tier1                               1900  1900     0     0
   jtreg:test/langtools:tier1                         3941  3941     0     0
   jtreg:test/nashorn:tier1                              0     0     0     0
   jtreg:test/jaxp:tier1                                 0     0     0     0
==============================
TEST SUCCESS
```

### gtest - PASSED

```
==============================
Test summary
==============================
   TEST                                              TOTAL  PASS  FAIL ERROR
   gtest:all/server                                    507   507     0     0
==============================
TEST SUCCESS
```

### GHA - RUNNING

https://github.com/TheMangovnik/jdk11u-dev/actions/runs/26746034005

---------

---------
### Progress
- \[ \] [JDK-8244336](https://bugs.openjdk.org/browse/JDK-8244336) needs maintainer approval
- \[x\] Change requires CSR request [JDK-8372102](https://bugs.openjdk.org/browse/JDK-8372102) to be approved
- \[ \] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issues
 * [JDK-8244336](https://bugs.openjdk.org/browse/JDK-8244336): Restrict algorithms at JCE layer (**Enhancement** - P3)
 * [JDK-8372102](https://bugs.openjdk.org/browse/JDK-8372102): Restrict algorithms at JCE layer (**CSR**)

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**) 🔄 Re-review required (review applies to [b56b2398](https://git.openjdk.org/jdk11u-dev/pull/3213/files/b56b2398264f097b9cd8bca16d416f4466684b90))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk11u-dev.git pull/3213/head:pull/3213` \
`$ git checkout pull/3213`

`$ git checkout pull/3213` \
`$ git pull https://git.openjdk.org/jdk11u-dev.git pull/3213/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3213`

`$ git pr show -t 3213`

</details>

<a href="https://git.openjdk.org/jdk11u-dev/pull/3213.diff">https://git.openjdk.org/jdk11u-dev/pull/3213.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk11u-dev/pull/3213#issuecomment-4591532590)
</details>

### DIFF
--- a/src/java.base/share/classes/javax/crypto/Cipher.java
+++ b/src/java.base/share/classes/javax/crypto/Cipher.java
@@ -529,7 +529,8 @@ public class Cipher {
 
         // throws NoSuchAlgorithmException if java.security disables it
         if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
-            throw new NoSuchAlgorithmException(transformation + " is disabled");
+            throw new NoSuchAlgorithmException(transformation +
+                " is disabled");
         }
 
         List<Transform> transforms = getTransforms(transformation);
--- a/src/java.base/share/classes/javax/crypto/Cipher.java
+++ b/src/java.base/share/classes/javax/crypto/Cipher.java
@@ -530,7 +530,7 @@ public class Cipher {
         // throws NoSuchAlgorithmException if java.security disables it
         if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
             throw new NoSuchAlgorithmException(transformation +
-                " is disabled");
+                    " is disabled");
         }
 
         List<Transform> transforms = getTransforms(transformation);
@@ -684,7 +684,8 @@ public class Cipher {
 
         // throws NoSuchAlgorithmException if java.security disables it
         if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
-            throw new NoSuchAlgorithmException(transformation + " is disabled");
+            throw new NoSuchAlgorithmException(transformation +
+                    " is disabled");
         }
 
         Exception failure = null;
--- a/src/java.base/share/classes/javax/crypto/Cipher.java
+++ b/src/java.base/share/classes/javax/crypto/Cipher.java
@@ -30,7 +30,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.regex.*;
 
-
 import java.security.*;
 import java.security.Provider.Service;
 import java.security.spec.AlgorithmParameterSpec;
@@ -45,6 +44,7 @@ import java.nio.ReadOnlyBufferException;
 
 import sun.security.util.Debug;
 import sun.security.jca.*;
+import sun.security.util.CryptoAlgorithmConstraints;
 
 /**
  * This class provides the functionality of a cryptographic cipher for
@@ -314,6 +314,7 @@ public class Cipher {
         if (transformation == null) {
             throw new NoSuchAlgorithmException("No transformation given");
         }
+
         /*
          * array containing the components of a Cipher transformation:
          *
@@ -525,6 +526,12 @@ public class Cipher {
         if ((transformation == null) || transformation.equals("")) {
             throw new NoSuchAlgorithmException("Null or empty transformation");
         }
+
+        // throws NoSuchAlgorithmException if java.security disables it
+        if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
+            throw new NoSuchAlgorithmException(transformation + " is disabled");
+        }
+
         List<Transform> transforms = getTransforms(transformation);
         List<ServiceId> cipherServices = new ArrayList<>(transforms.size());
         for (Transform transform : transforms) {
@@ -673,6 +680,12 @@ public class Cipher {
         if (provider == null) {
             throw new IllegalArgumentException("Missing provider");
         }
+
+        // throws NoSuchAlgorithmException if java.security disables it
+        if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
+            throw new NoSuchAlgorithmException(transformation + " is disabled");
+        }
+
         Exception failure = null;
         List<Transform> transforms = getTransforms(transformation);
         boolean providerChecked = false;
--- a/src/java.base/share/classes/sun/security/util/AbstractAlgorithmConstraints.java
+++ b/src/java.base/share/classes/sun/security/util/AbstractAlgorithmConstraints.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -50,7 +50,16 @@ public abstract class AbstractAlgorithmConstraints
 
     // Get algorithm constraints from the specified security property.
     static Set<String> getAlgorithms(String propertyName) {
-        String property = AccessController.doPrivileged(
+        return getAlgorithms(propertyName, false);
+    }
+
+    // Get algorithm constraints from the specified security property or
+    // system property if allowSystemOverride == true.
+    static Set<String> getAlgorithms(String propertyName,
+            boolean allowSystemOverride) {
+        String property = allowSystemOverride ?
+            SecurityProperties.privilegedGetOverridable(propertyName) :
+            AccessController.doPrivileged(
                 new PrivilegedAction<String>() {
                     @Override
                     public String run() {
@@ -75,7 +84,8 @@ public abstract class AbstractAlgorithmConstraints
         if (algorithmsInProperty == null) {
             return Collections.emptySet();
         }
-        Set<String> algorithmsInPropertySet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        Set<String> algorithmsInPropertySet = new
+                TreeSet<>(String.CASE_INSENSITIVE_ORDER);
         algorithmsInPropertySet.addAll(Arrays.asList(algorithmsInProperty));
         return algorithmsInPropertySet;
     }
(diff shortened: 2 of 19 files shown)


src/java.base/share/classes/javax/crypto/Cipher.java
```diff
@@ -525,6 +526,12 @@ public static final Cipher getInstance(String transformation)
         if ((transformation == null) || transformation.equals("")) {
             throw new NoSuchAlgorithmException("Null or empty transformation");
         }
+
+        // throws NoSuchAlgorithmException if java.security disables it
+        if (!CryptoAlgorithmConstraints.permits("Cipher", transformation)) {
+            throw new NoSuchAlgorithmException(transformation + " is disabled");
```
jerboaa: Nit: The JDK 17u patch has this on two lines:

```suggestion
            throw new NoSuchAlgorithmException(transformation +
                " is disabled");
```

src/java.base/share/classes/sun/security/util/AbstractAlgorithmConstraints.java
```diff
@@ -75,7 +84,8 @@ public String run() {
         if (algorithmsInProperty == null) {
             return Collections.emptySet();
         }
-        Set<String> algorithmsInPropertySet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        Set<String> algorithmsInPropertySet = new
+                TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```
jerboaa: Nit: Odd formatting that is different to the JDK 17u patch:

```suggestion
        Set<String> algorithmsInPropertySet =
               new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```

jerboaa (approved):
This looks OK to me. Two nits.

jerboaa (approved):
OK

gnu-andrew (changes requested):
* Differences in `Keystore.java` are due to the absence of [JDK-8245679](https://bugs.openjdk.org/browse/JDK-8245679) and are adapted appropriately.  8245679 itself looks a risky backport as it changes Provider lookup behaviour.
* I don't really see the need to move the declaration of `MessageDigest md` in either patch, but at least in 17, it is moved after the new block which could cause the method to exit early.  I don't see any reason to move it in 11u as the context is quite different.
* `MessageDigest.java` has a missing newline in 11u compared with 17u before `if (provider == null)`
* There are differences in `Signature.java` but this is in erased code so no issue
* `Cipher.java` still contains differences in the indentation of `NoSuchAlgorithmException` construction as pointed out by Severin.  The first where correction was attempted is not indented correctly.  The second instance has not been corrected at all.
* `AbstractAlgorithmConstraints.java` also has bad indentation of the `TreeSet` line raised by Severin.  The first part of the patch also looks different, but this is because 11u never had the `SuppressWarnings` annotation added; changes are otherwise the same.
* In `CryptoAlgorithmConstraints.java`, the correction to `debug.println` should be a change to `Debug.println` rather than changing the arguments.  The method exists in 11u but is static, rather than an instance method.  Changing it to call the static version will mean that, if the dual argument method is changing to an instance method at a later date, this will be caught by a compile error.  If you change the arguments as in the current patch, this difference will remain unnoticed.  The change is made by [JDK-8051959](https://bugs.openjdk.org/browse/JDK-8051959) which adds thread info and timestamps to the output.  This has already been proposed for 11u, but was blocked by the need for an earlier change, so I suspect this is likely to be integrated at some point.

You can see the differences I'm referring to by running `diff -u` on your patch vs. [the 17u patch](https://github.com/openjdk/jdk17u/commit/0c5d6e1768ed40c86a05122672cad59cebe005fe).  I would strongly suggest doing this for backports so you are aware of differences and can catch ones that shouldn't be there before submitting the patch for review.

TheMangovnik:
> * Differences in `Keystore.java` are due to the absence of [JDK-8245679](https://bugs.openjdk.org/browse/JDK-8245679) and are adapted appropriately.  8245679 itself looks a risky backport as it changes Provider lookup behaviour.

👍

---

> * I don't really see the need to move the declaration of `MessageDigest md` in either patch, but at least in 17, it is moved after the new block which could cause the method to exit early.  I don't see any reason to move it in 11u as the context is quite different.

tl;dr: I see the move of the `MessageDigest md`, but I do not see how it is different between the jdk11 and jdk17 patch.

I have been peering into the patches, both in jdk17 and in this PR, and I fail to fully understand what you are referencing.
- I see that in both patches declaration of `MessageDigest md` moved in file `MessageDigest.java`.
- I see that in both patches this occurred in the method `public static MessageDigest getInstance(String algorithm)`
- I see that in both patches declaration of `MessageDigest md` swapped places with the already present command below it. Functionally doing what is in the following code block. I remember looking at this when doing this backport and wondering about this change too, but I have decided to keep this in since it should not change anything anyway.
```diff
- MessageDigest md;
Object[] objs = Security.getImpl(algorithm, "MessageDigest",
                                             (String)null);
+ MessageDigest md;
```
- I see that in both patches the declaration of `MessageDigest md` is after the newly introduced block, resulting in the code snipped below in both jdk17 and jdk11:
```java
if (!CryptoAlgorithmConstraints.permits("MessageDigest", algorithm)) {
     throw new NoSuchAlgorithmException(algorithm + " is disabled");
}

// In this place also starts a try block in jdk11. But that should be inconsequential for this situation.

Object[] objs = Security.getImpl(algorithm, "MessageDigest",
                                             (String)null);
MessageDigest md;
```
- So as far as I understand -> the declaration is always after the new block, the move of the declaration is relative to command below it. If I understand your comment correctly, you are saying that the previous statement is not true?

I see similar castling in the method overload just below it `public static MessageDigest getInstance(String algorithm, String provider)`. In jdk11 patch there is no change in the position of the declaration of `MessageDigest md`, since the declaration itself lives in the if block in this method. However there is same position change of the declaration in jdk17 patch, because the declaration is factored out of the if block. But even here the declaration is after the newly introduced block in all cases.

---

> * `MessageDigest.java` has a missing newline in 11u compared with 17u before `if (provider == null)`

Added.

---

> * There are differences in `Signature.java` but this is in erased code so no issue

👍

---

> * `Cipher.java` still contains differences in the indentation of `NoSuchAlgorithmException` construction as pointed out by Severin.  The first where correction was attempted is not indented correctly.  The second instance has not been corrected at all.

This is stupid oversight from my side. Corrected.

---

> * `AbstractAlgorithmConstraints.java` also has bad indentation of the `TreeSet` line raised by Severin.  The first part of the patch also looks different, but this is because 11u never had the `SuppressWarnings` annotation added; changes are otherwise the same.

Indentation corrected.

---

> * In `CryptoAlgorithmConstraints.java`, the correction to `debug.println` should be a change to `Debug.println` rather than changing the arguments.  The method exists in 11u but is static, rather than an instance method.  Changing it to call the static version will mean that, if the dual argument method is changing to an instance method at a later date, this will be caught by a compile error.  If you change the arguments as in the current patch, this difference will remain unnoticed.  The change is made by [JDK-8051959](https://bugs.openjdk.org/browse/JDK-8051959) which adds thread info and timestamps to the output.  This has already been proposed for 11u, but was blocked by the need for an earlier change, so I suspect this is likely to be integrated at some point.

I see, the fact that the `Debug.println()` is within a if block that checks for an instance of the Debug class confused me here. Nevertheless, it is corrected to the static variant `Debug.println()` now.

---
 
> You can see the differences I'm referring to by running `diff -u` on your patch vs. [the 17u patch](https://github.com/openjdk/jdk17u/commit/0c5d6e1768ed40c86a05122672cad59cebe005fe). I would strongly suggest doing this for backports so you are aware of differences and can catch ones that shouldn't be there before submitting the patch for review.

I will keep this in mind and do more throughout job next time.

jerboaa:
@TheMangovnik Please reopen the PR. You can use `/open`.

